PR #165, opened 2017-08-18
Pick better fallback aliases for import references by using their 'namespace object paths'.

bicknellr:
Fixes #108.


bicknellr:
Squashed. Also, I think it might actually make sense to leave `generateAliases` (now `generateMemberPathAliases`) where it is so that `generateAliasRequests` doesn't have to contain all the logic to determine what aliases are best for a particular `ImportReference`.

src/document-converter.ts
```diff
@@ -803,12 +829,17 @@ export class DocumentConverter {
             return;
           }
           const [callPath] = assignmentPath.replace(jsc.callExpression(
-              jsc.identifier(setterName), [assignmentPath.node.right]));
+              jsc.identifier('IMPORT_REFERENCE_REPLACEMENT_FAILED'),
```
FredKSchott: Handle this error (throw an exception? log a warning?) instead of replacing the node with a dummy identifier

src/document-converter.ts
```diff
@@ -1085,6 +1120,15 @@ export class DocumentConverter {
     }
     const usedIdentifiers = collectIdentifierNames(program, ignoredIdentifiers);
 
+    const allImportReferences = new Set();
+    for (const refs of importedReferences.values()) {
+      for (const ref of refs) {
```
FredKSchott: code ⛳️ : `allImportReferences.add(...refs)` lets you skip the nested for-loop

src/document-converter.ts
```diff
@@ -1284,3 +1336,43 @@ function domModuleCanBeInlined(domModule: parse5.ASTNode) {
...
+    undesirableAliases: ReadonlySet<string>): Map<JsExport, string[]> {
+  const aliasRequests = new Map<JsExport, Array<string>>();
+  for (const {target, requestedIdentifiers} of importReferences) {
+    let aliasRequestsForImport = aliasRequests.get(target);
+    if (aliasRequestsForImport === undefined) {
+      aliasRequestsForImport = [];
+      aliasRequests.set(target, aliasRequestsForImport);
+    }
+
+    for (const name of requestedIdentifiers) {
```
FredKSchott: same here: `aliasRequestsForImport.push(... requestedIdentifiers);` is cleaner

src/document-converter.ts
```diff
@@ -1284,3 +1336,43 @@ function domModuleCanBeInlined(domModule: parse5.ASTNode) {
 
   return true;
 }
+
+/**
+ * Produces an array of preferred aliases for all `JsExports` referenced by a
+ * set of `ImportReference`s, given a set of undesirable aliases.
+ */
+function generateAliasRequests(
```
FredKSchott: I'm a little concerned with the perf for this function, considering it's run for every script.

src/document-converter.ts
bicknellr: The identifier here should always be replaced but I have to supply some name for it at this point so I figured it would be good to give it a name that made it obvious that it shouldn't be in the output.

src/document-converter.ts
bicknellr: I don't think `add` accepts multiple arguments, does it?

src/document-converter.ts
bicknellr: (Inserting this identifier into the program at this point is not the potential error that this identifier's name is meant to bring attention to.)

src/document-converter.ts
FredKSchott: bah, you're totally right. nvm!

src/document-converter.ts
FredKSchott: But does this get replaced again later? Otherwise, aren't you silently adding an invalid identifier to your code?

If this is absolutely required, at the very least please add some documentation to why this is okay & where it is handled later in the flow.

bicknellr: > But does this get replaced again later?

Yes, this identifier is [set as the AST node (`path`) for this `ImportReference`](https://github.com/Polymer/polymer-modulizer/pull/165/files#diff-4c7c2c6e37f696bf962a9f9da25cf010R840), so it should always be replaced later, during import aliasing. However, this particular point in the modulizer code is where some gets/sets are transformed into calls to modulizer-created and imported getter/setter functions: I'm not really sure how it determines when to do this, but modulizer seems to detect if you were trying to do something like `Scope.x = y;` and changes it to something like `import { setX } from "./Scope.js"; /* ... */ setX(y);`, if appropriate. So, AFAICT, I have to produce an AST node for the 'function' part (`setX`) of the 'function call' node (the whole `setX(y)`) even though this node is also the one that will be replaced during aliasing and shouldn't ever appear in the output.

You're totally right about this needing some documentation. I'll add a comment here to explain why this is happening.

bicknellr: I added the comment, let me know if you think it seems clear enough.

src/document-converter.ts
bicknellr: (This function has moved into 'src/document-util.ts', but isn't changed.)

bicknellr:
This code is overly cautious in that it assumes that any identifier in the same module should be considered conflicting when assigning import aliases. A couple of cases that I think I should fix before trying to land this are (1) identifiers in the tail of a chain of dot operator applications (`a.b.c.d`: all identifiers except `a`) and (2) identifiers in exports (still need to figure out if this is always true).

(1) causes something that could be this:
```
import {a} from "./module0.js";
import {b} from "./module1.js";
a.b();
b();
```
to end up as:
```
import {a} from "./module0.js";
import {b$0} from "./module1.js";
a.b();
b$0();
```
even though `b` doesn't conflict.

(2) causes what could be this:
```
import {a} from "./module.js";
export {a};
```
to be:
```
import {a as a$0} from "./module.js";
export {a$0 as a};
```
even though `a` doesn't conflict.

justinfagnani:
At some point I think we want to change the parser to Babylon which has scope analysis that should be able to tell us relatively easily if there's a naming conflict. Alternatively, since the Analyzer 3.x already uses Babylon, we just write the scan passes as actual Analyzer scanners.

PR #189, opened 2017-08-26
New: Remove top-level strict mode directive

### LINKED ISSUE #181: Should we remove 'use-strict'?
Should this remove the `'use strict';` declaration since it is always enabled in modules?

stramel:
Appears that we were removing it in IIFEs already but not toplevel. Fixes #181 

Can add tests if you guys are good with this.


stramel:
Ping @rictic

stramel:
@rictic @FredKSchott  @usergenic Do you guys have any feedback on this?

PR #190, opened 2017-08-26
New: Check for dirty repository

stramel:
Check that the repository is clean before running instead of requiring users to use `--force` flag.

Open to feedback :)

Can add tests if you guys are okay with this approach.


stramel:
Ping @rictic

stramel:
Going to close this one as the mainline has gone a different route from nodegit. Will try to get around to re-implementing this but if someone has time, feel free to take over.

PR #192, opened 2017-08-30
Preserve comments in HTML->JS transform.

rictic:
Fixes https://github.com/Polymer/polymer-modulizer/issues/191

### DIFF
--- a/src/document-converter.ts
+++ b/src/document-converter.ts
@@ -189,6 +189,7 @@ export class DocumentConverter {
       }
     }
     const combinedToplevelStatements = [];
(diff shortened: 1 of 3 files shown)


src/document-converter.ts
```diff
@@ -1284,3 +1295,115 @@ function domModuleCanBeInlined(domModule: parse5.ASTNode) {
 
   return true;
 }
+
+/**
+ * Yields all nodes inside the given node in top-down, first-to-last order.
```
justinfagnani: "in-order"?

PR #204, opened 2017-09-21
Refactor workspace logic into polymer-workspaces standalone library

FredKSchott:
#### 💨 polymer-workspaces: installs ~98% faster without buggy nodegit compilation 
(311.61s -> 7.71s)
#### 💄 Prettier UI

screen shot 2017-10-05 at 2 36 04 pm


src/cli.ts
```diff
@@ -132,9 +110,16 @@ const optionDefinitions: commandLineArgs.OptionDefinition[] = [
         `If given, may overwrite or delete files when converting the given ` +
         `input directory.`,
   },
+  {
+    name: 'fresh',
```
justinfagnani: 🚲 🏠 time: how about `clean`? It's pretty much the standard term for this.

src/cli/command-workspace.ts
```diff
@@ -0,0 +1,100 @@
...
+echo 'PASTE TOKEN HEX HERE' > ./github-token
+`;
+
+
+/**
+ * Checks for github-token in the RunnerOptions and if not specified, will look
+ * in the github-token file in working folder.  If that doesn't exist either,
+ * we message to the user that we need a token and exit the process.
+ */
+function loadGitHubToken(): string|null {
```
justinfagnani: seem like this could move into polymer-workspaces

justinfagnani:
@FredKSchott I'm giving this another review now, in case you were going to merge, hold off a sec

src/cli/command-workspace.ts
FredKSchott: I like that the library itself is agnostic to how the token is loaded. That makes it flexible enough for scripts, CLIs, etc.

It could be interesting as a helper function though, to at least make it easier to build CLIs.

PR #210, opened 2017-10-10
Refactor the CLI logic interface to support both package layouts

FredKSchott:
***11/17: Updated for a post-layout-refactoring world! ™***

The CLI currently runs both single-package & multi-package (workspace) through the same CLI interface, with different options triggering both. `modulizer` will normally run in single-package mode, but the `--repo` turns on the completely different workspace workflow. `--repo` acts less like an option, and more like a command.

This PR refactors the CLI arguments logic with the addition of a new explicit conversion "mode". Instead of configuring package vs. workspace through some combination of flags, the user now sets the mode explicitly via a command:

```
$ modulizer package --out . --force
$ modulizer workspace --repo PolymerElements/paper-*#2.0-preview --clean
```

### DIFF
--- a/src/cli/command-help.ts
+++ b/src/cli/command-help.ts
@@ -1,0 +1,37 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {CliOptions, commandLineArgDefinitions} from './args';
+
+
+export default async function run(_options: CliOptions) {
+  const getUsage = require('command-line-usage');
+  const usage = getUsage([
+    {
+      header: 'modulizer',
+      content: `Convert HTML Imports to JavaScript modules
+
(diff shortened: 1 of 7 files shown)


justinfagnani:
I'm not sure we can really make the distinction between element and application. Aside from dealing with absolute vs relative URLs, what's the difference?

FredKSchott:
@justinfagnani @usergenic this has been rebased off of master, and removes the distinction of element vs. package at the top level (with the plan to get that removed throughout the codebase as well).

PTAL

src/cli/index.ts
```diff
@@ -12,175 +12,42 @@
...
+  switch (options.mode) {
+    case 'package':
+      return runPackageCommand(options);
+
+    case 'workspace':
+      return runWorkspaceCommand(options);
+
+    default:
+      console.error(`No mode provided! Run "modulizer [mode] [options...]"`);
+      console.error(`Available modes: element, workspace`);
```
stramel: This states element and workspace but above it says package and workspace

src/cli/command-help.ts
```diff
@@ -0,0 +1,37 @@
...
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {CliOptions, commandLineArgDefinitions} from './args';
+
+
+export default async function run(_options: CliOptions) {
```
stramel: Is `_options` used in here?

justinfagnani:
@FredKSchott rebase to get this one going again?

justinfagnani:
@FredKSchott if we remove package mode, we'll close this, yeah?

usergenic:
Is this PR still a thing?

PR #215, opened 2017-11-01
New: Check the directory git status

### LINKED ISSUE #135: Add guard rails
We should require that the directory either be a clean git repo or the user pass in a `--hurt-me-plenty` flag in order to run in non-workspace mode.

stramel:
Follow up to #190 
Fixes #135 

Unfortunately, @rictic, I didn't impement the `--hurt-me-plenty` flag. Let me know if you still want it :laughing: 

Checks the `in` directory for anything changes in git.  Must run with `--force` if you want to run anyways.


stramel:
Ping @rictic @FredKSchott

stramel:
@FredKSchott @usergenic @rictic Rebased onto the latest changes in master

src/cli/command-package.ts
```diff
@@ -22,14 +23,27 @@ import {readJson} from '../manifest-converter';
...
-  // for now, we'll ask the user to pass in a --force flag.
-  if (!options.force) {
-    console.error(
-        `When running modulizer on an existing directory, ` +
-        `be sure that all changes are checked into source control. ` +
-        `Run with --force once you've verified.`);
-    process.exit(1);
+  // We need to be sure it's safe.
+  try {
+    await new Promise((resolve, reject) => {
```
FredKSchott: That's funny, I needed this for the fixture tests as well. I ended up implementing a wrapper interface that's been really useful in polymer-workspaces: https://github.com/Polymer/polymer-modulizer/blob/6791dd4ba7ee6960a021d193219879cfa8675d21/src/test/util/exec.ts

Want to bring that file in to your PR and then use it here?

src/cli/command-package.ts
```diff
@@ -22,14 +23,27 @@ import {readJson} from '../manifest-converter';
...
+      exec('git status -s', {
+        cwd: options.in || process.cwd()
+      }, (error, stdout, stderr) => {
+        if (error || stderr || stdout) {
+          return reject();
```
FredKSchott: If an actual `error` happens it's probably not a git clean vs. unclean issue. Reject the error object and report it somehow below in the catch.

src/cli/command-package.ts
```diff
@@ -22,14 +23,27 @@ import {readJson} from '../manifest-converter';
...
-    console.error(
-        `When running modulizer on an existing directory, ` +
-        `be sure that all changes are checked into source control. ` +
-        `Run with --force once you've verified.`);
-    process.exit(1);
+  // We need to be sure it's safe.
+  try {
+    await new Promise((resolve, reject) => {
+      exec('git status -s', {
+        cwd: options.in || process.cwd()
```
FredKSchott: Move up the `inDir` variable defined below and then reference it here instead

src/cli/command-package.ts
```diff
@@ -22,14 +23,27 @@ import {readJson} from '../manifest-converter';
...
+          return reject();
+        }
+
+        resolve();
+      });
+    });
+  } catch (ex) {
+    if (!options.force) {
+      console.error(
+          `When running modulizer on an existing directory, ` +
```
FredKSchott: nit: The user should know explicitly what went wrong. Reword to something like:

"Git repo is dirty. Check all changes in to source control and then try again."

src/cli/command-package.ts
stramel: The only `error` I was noticing was that the command wasn't found. We could either let this throw or do the `checkCommand` function and handle it. I'm not sure how you guys want to handle that case so I just left it as throwing for now.

src/test/util/exec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export interface ExecResult {
+  stdout: string;
+  stderr: string;
+}
+
+/**
+ * A helper function for working with Node's core execFile() method.
+ */
+export default async function exec(
```
stramel: @FredKSchott Should I move these into the `src/util.ts` file instead of under test utils?

src/cli/command-package.ts
stramel: Thanks, this definitely cleaned up the implementation. :+1: Should I move it to `src/util.ts` instead of `src/test/util/exec.ts`

src/test/util/exec.ts
FredKSchott: 👍 that makes sense, I'll make sure I get that upstream in my PR

src/cli/command-package.ts
FredKSchott: Sorry, I meant that I just meant to make sure a true error wasn't swallowed and reported under "hey your git repo is dirty". `reject(err)` and then doing something with the caught error would work.

The new change accomplishes this though too. If `exec()` rejects, the true error message will halt and bubble up.

src/util.ts
```diff
@@ -262,3 +274,36 @@ export async function writeFileResults(
...
+    err.cwd = cwd;
+    throw err;
+  }
+}
+
+/**
+ * A helper function to check whether a shell command exists or not.
+ * Useful for user-facing warnings.
+ */
+export async function checkCommand(commandName: string): Promise<boolean> {
```
FredKSchott: no need for this (I don't see it used anyway). It will already have been internally inside of polymer-workspaces, and even still the error thrown by exec is fine for our use-case.

stramel: Good point. Originally transferred when I was referencing your file to prevent more potential conflicts.

FredKSchott: Ha, I guess I made the same mistake 😛

src/util.ts
```diff
@@ -22,21 +23,32 @@ import * as parse5 from 'parse5';
...
 
+export interface ExecResult {
+  stdout: string;
+  stderr: string;
+}
+
+/**
+ * Helper promisified "execFile" library function.
+ */
+export const execFile = promisify(_execFile);
```
FredKSchott: The `exec()` helper below should be the only wrapper we expose. No worries about making another change, I'll fix this on merge.

PR #216, opened 2017-11-02
[ProjectConverter] Fix outputs keying to prevent cache missing, duplicates

FredKSchott:
*First PR! Start here!*

## Summary:

Previously, `outputs` contained a mishmash of url. Some used the old bower file location, some used the new npm file location, and some were manually constructed. This has caused cache misses and slower conversions (and annoying bugs as I've been poking around and refactoring).

This PR fixes all of these issues by always keying to the OriginalDocumentUrl, along with some naming clarifications.

### DIFF
--- a/src/base-converter.ts
+++ b/src/base-converter.ts
@@ -134,27 +134,44 @@ export abstract class BaseConverter {
   }
 
   /**
+   * Check if a document is explicitly excluded or has already been converted
+   * to decide if it should be converted or skipped.
+   */
+  shouldConvertDocument(document: Document): boolean {
+    const documentUrl = getDocumentUrl(document);
+    if (this.results.has(documentUrl)) {
+      return false;
+    }
+    for (const exclude of this.excludes) {
+      if (documentUrl.endsWith(exclude)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
    * Converts a Polymer Analyzer HTML document to a JS module
    */
   convertDocument(document: Document, visited: Set<OriginalDocumentUrl>) {
-    if (!this.results.has(getDocumentUrl(document))) {
-      const newModule =
-          this.getDocumentConverter(document, visited).convertToJsModule();
-      if (newModule !== undefined) {
-        this.handleNewJsModules(newModule);
-      }
+    if (!this.shouldConvertDocument(document)) {
+      return;
     }
+    const newModule =
+        this.getDocumentConverter(document, visited).convertToJsModule();
+    this.handleNewJsModules(newModule);
   }
 
   /**
    * Converts HTML Imports and inline scripts to module scripts as necessary.
    */
   convertHtmlToHtml(document: Document, visited: Set<OriginalDocumentUrl>) {
-    if (!this.results.has(getDocumentUrl(document))) {
-      const newModule = this.getDocumentConverter(document, visited)
-                            .convertAsToplevelHtmlDocument();
-      this.handleNewJsModules(newModule);
+    if (!this.shouldConvertDocument(document)) {
+      return;
     }
+    const newModule = this.getDocumentConverter(document, visited)
+                          .convertAsToplevelHtmlDocument();
+    this.handleNewJsModules(newModule);
   }
 
   private handleNewJsModules(newModule: ConversionResult): void {
--- a/src/base-converter.ts
+++ b/src/base-converter.ts
@@ -140,7 +140,7 @@ export abstract class BaseConverter {
     if (!this.results.has(getDocumentUrl(document))) {
       const newModule =
           this.getDocumentConverter(document, visited).convertToJsModule();
-      if (newModule) {
+      if (newModule !== undefined) {
         this.handleNewJsModules(newModule);
       }
     }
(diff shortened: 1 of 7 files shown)


src/tools/test-polymer.ts
```diff
@@ -75,7 +75,7 @@ function rework(line: string) {
                           .filter(([_, v]) => v !== undefined)
                           .map(([k]) => k);
   const expectedPaths = IterableX.from(walkDir(expectedDir))
-                            .map((f) => `./${f}` as ConvertedDocumentUrl)
+                            .map((f) => `${f}` as ConvertedDocumentFilePath)
```
usergenic: Immaterial, but style-wise it feels like, for consistency, this should be:
```ts
.map((f) => f as string as ConvertedDocumentFilePath)
```

FredKSchott: good catch, no longer need the template string since there's no more string modifications

FredKSchott:
Thanks @usergenic! @justinfagnani PTAL

src/document-converter.ts
```diff
@@ -235,24 +235,22 @@ export class DocumentConverter {
...
         recast.print(program, {quote: 'single', wrapColumn: 80, tabWidth: 2});
-    const deletedHtmlFileUrl =
-        ('./' + this.originalUrl) as ConvertedDocumentUrl;
-    return [
-      {
+
+    return {
+      originalUrl: this.originalUrl,
+      convertedUrl: this.convertedUrl,
+      convertedFilePath: this.originalUrl.replace('.html', '.js') as
```
justinfagnani: This looks potentially dangerous due to the unrestricted replace. Can you move to a central method to convert .html URLs to .js, and maybe only replace extensions?

src/document-converter.ts
```diff
@@ -182,10 +182,10 @@ export class DocumentConverter {
         });
   }
 
-  convertToJsModule(): Iterable<ConversionOutput> {
+  convertToJsModule(): ConversionResult|void {
     for (const exclude of this.conversionMetadata.excludes) {
       if (this.originalUrl.endsWith(exclude)) {
-        return [];
+        return;
```
justinfagnani: Would it make sense to return a no-op ConversionResult? You use `results.has()` as a gate above, and populating the map would prevent multiple visits to this method.

src/document-converter.ts
```diff
@@ -404,11 +402,17 @@ export class DocumentConverter {
...
-    return [{
-      type: 'html-file',
-      url: ('./' + this.originalUrl) as ConvertedDocumentUrl,
-      source: contents,
-    }];
+
+    return {
+      originalUrl: this.originalUrl,
+      convertedUrl: this.convertedUrl,
+      convertedFilePath: this.originalUrl as string as
```
justinfagnani: Consider making a no-op conversion method that just does the cast.

src/js-module.ts
```diff
@@ -12,42 +12,31 @@
...
+  /** Set of exported names. */
   readonly es6Exports: ReadonlySet<string>;
+}
 
-  readonly exportedNamespaceMembers: ReadonlyArray<NamespaceMemberToExport>;
+export interface ConversionResult {
+  readonly originalUrl: OriginalDocumentUrl;
+  readonly convertedUrl: ConvertedDocumentUrl;
+  readonly convertedFilePath: ConvertedDocumentFilePath;
+  readonly keepOriginal?: boolean;
```
justinfagnani: Can you document this one? What are the implied side effects of this being true? What's the intended default?

src/base-converter.ts
```diff
@@ -112,63 +113,65 @@ export abstract class BaseConverter {
...
-  private handleNewJsModules(outputs: Iterable<ConversionOutput>): void {
-    for (const output of outputs) {
-      this.outputs.set(output.url, output);
-      if (output.type === 'js-module') {
-        for (const expr of output.exportedNamespaceMembers) {
-          this.namespacedExports.set(
-              expr.oldNamespacedName,
-              new JsExport(output.url, expr.es6ExportName));
-        }
+  private handleNewJsModules(newModule: ConversionResult): void {
```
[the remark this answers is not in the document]
FredKSchott: This interface is superseded by `UrlHandler`: #224

src/document-converter.ts
FredKSchott: Good idea, I must have missed this when I was looking for other instances of fixHtmlExtension. Moving this into a helper method with the correct implementation.

src/base-converter.ts
```diff
@@ -112,63 +113,65 @@ export abstract class BaseConverter {
...
    */
   convertHtmlToHtml(document: Document, visited: Set<OriginalDocumentUrl>) {
-    const htmlUrl = `./${document.url}` as ConvertedDocumentUrl;
-    if (!this.outputs.has(htmlUrl)) {
-      const newModules = this.getDocumentConverter(document, visited)
-                             .convertAsToplevelHtmlDocument();
-      this.handleNewJsModules(newModules);
+    if (!this.results.has(getDocumentUrl(document))) {
+      const newModule = this.getDocumentConverter(document, visited)
+                            .convertAsToplevelHtmlDocument();
```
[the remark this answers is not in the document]
FredKSchott: No reason that I can tell. I'll move this check a level up so that this method always performs the conversion its told to.

FredKSchott:
@justinfagnani PTAL

PR #219, opened 2017-11-07
[ProjectConverter] Clean up CLI setup/layout

FredKSchott:
*Builds off of #216: "Fix outputs keying to prevent cache missing, duplicates"*

## Summary:

This PR cleans up some of the CLI logic and partially-moved file structure.

- some bug fixes around path resolution for paths provided by the CLI
- The main CLI entrypoint is moved into the `cli/` subdirectory with the other CLI logic
- inDir & outDir are now always resolved by the time they get to the `convert()` method.
- workspace conversion runs off the polymer-workspaces list of repos, instead of a directory scan

### DIFF
--- a/src/convert-package.ts
+++ b/src/convert-package.ts
@@ -12,16 +12,13 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import * as fs from 'mz/fs';
-import * as path from 'path';
 import {Analysis, Analyzer, FSUrlLoader, InMemoryOverlayUrlLoader, PackageUrlResolver} from 'polymer-analyzer';
-import * as rimraf from 'rimraf';
 
 import {AnalysisConverter, AnalysisConverterOptions} from './analysis-converter';
 import {generatePackageJson, readJson, writeJson} from './manifest-converter';
 import {polymerFileOverrides} from './special-casing';
+import {mkdirp, rimraf, writeFileResults} from './util';
 
-const mkdirp = require('mkdirp');
 
 
 type ConvertPackageOptions = AnalysisConverterOptions&{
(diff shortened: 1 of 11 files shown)


src/convert-package.ts
```diff
@@ -50,8 +50,39 @@ type ConvertPackageOptions = AnalysisConverterOptions&{
...
+
+/**
+ * Create and/or clean the "out" directory, setting it up for conversion.
+ */
+async function setupOutDir(outDir: string, clean: boolean) {
+  if (clean) {
+    rimraf.sync(outDir);
+  }
+  try {
+    await fs.mkdir(outDir);
```
justinfagnani: mkdirp does this. no?

src/convert-package.ts
```diff
@@ -86,51 +117,33 @@ export function configureConverter(
...
   for (const [newPath, newSource] of results) {
-    const outPath = path.resolve(outDirResolved, newPath);
-    mkdirp.sync(path.dirname(outPath));
-    if (newSource !== undefined) {
-      await fs.writeFile(outPath, newSource);
-    } else if (fs.existsSync(outPath)) {
-      await fs.unlink(outPath);
-    }
+    const filePath = path.resolve(outDir, newPath);
+    await writeFile(filePath, newSource);
```
justinfagnani: If you wanted to try to get some parallelism here (I'm curious if fs writes even benefit), you could use Promise.all:

```ts
await Promise.all(IterableX.from(results).map(([newPath, newSource]) => 
    writeFile(path.resolve(outDir, newPath), newSource)));
```

src/convert-package.ts
```diff
@@ -50,8 +50,39 @@ type ConvertPackageOptions = AnalysisConverterOptions&{
...
+      throw e;
+    }
+  }
+}
+
+/**
+ * Create and/or clean the "out" directory, setting it up for conversion.
+ */
+async function writeFile(outPath: string, newSource: string|undefined) {
+  mkdirp.sync(path.dirname(outPath));
```
justinfagnani: you could promisify this

FredKSchott: Yup, just realized convert-workspace is doing that

src/convert-package.ts
FredKSchott: Not sure if Node can do parallel writes but it can at least queue them up at I/O level more efficiently. This should have an even bigger effect on convert-workspace.

PR #220, opened 2017-11-07
[ProjectConverter] Clean up url-converter logic/layout

FredKSchott:
*Builds off of #219: "[ProjectConverter] Clean up CLI setup/layout"*

## Summary:

This PR sets up the larger refactoring work of #217 by cleaning up some of the URL handling logic.

- No behavior changes 
- Temporarily move url-converter logic into `urls/util.ts`, in preparation for final move to UrlHandlers
- Break out types into separate `urls/types.ts` file, in preparation for additional url-related types

### DIFF
--- a/src/urls/types.ts
+++ b/src/urls/types.ts
@@ -1,0 +1,35 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+/**
+ * A URL path to an document, pre-conversion. Always relative to the current
+ * project layout (package, workspace, etc).
+ */
+export type OriginalDocumentUrl = string&{_OriginalDocumentUrl: never};
+
+/**
+ * A URL path to a document, post-conversion. Uses npm naming for all URLs
+ * containing package names. Always relative to the current project layout
+ * (package, workspace, etc).
+ */
+export type ConvertedDocumentUrl = string&{_ConvertedDocumentUrl: never};
+
+/**
+ * A file path to where a document will be written, post-conversion. Unlike
+ * ConvertedDocumentUrl, this URL preserves the original package folder.
+ * Useful in workspace layouts where converted files should still be written
+ * to the original, analyzed repo.
+ */
+export type ConvertedDocumentFilePath =
+    string&{_ConvertedDocumentFilePath: never};
--- a/src/urls/types.ts
+++ b/src/urls/types.ts
@@ -13,7 +13,7 @@
  */
 
 /**
- * A URL path to an document, pre-conversion. Always relative to the current
+ * A URL path to a document, pre-conversion. Always relative to the current
  * project layout (package, workspace, etc).
  */
 export type OriginalDocumentUrl = string&{_OriginalDocumentUrl: never};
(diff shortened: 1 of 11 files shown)


src/urls/util.ts
```diff
@@ -14,32 +14,14 @@
...
- * ConvertedDocumentUrl, this url keeps the original package folder and Bower
- * package name. Useful in workspace projects where converted files should still
- * be written to the original analyzed repo.
- */
-export type ConvertedDocumentFilePath =
-    string&{_ConvertedDocumentFilePath: never};
+import {ConvertedDocumentUrl, OriginalDocumentUrl} from './types';
 
+/** The HTML file extension. */
+export const htmlExtension = '.html';
```
rictic: Do these make sense as constants? Is there any case where we'd expect them to be typo'd or that it would be clearer to have the name of the constant rather than just the string literal?

src/urls/types.ts
```diff
@@ -0,0 +1,35 @@
...
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+/**
+ * A URL path to an document, pre-conversion. Always relative to the current
```
rictic: `an document` -> `a document`

src/urls/util.ts
FredKSchott: That's a fair point, thanks for bringing it up. I don't feel too strongly either way here, but the one place that it definitely feels safer is here:

```ts
if (jsUrl.endsWith(htmlExtension)) {
  jsUrl = fixHtmlExtension(jsUrl) as ConvertedDocumentUrl;
}
```

Checking for the existence of a constant (`htmlExtension`) makes me feel more confident calling the method built to handle that constant (`fixHtmlExtension()`). But digging deeper, that's really the result of `fixHtmlExtension()` breaking a non-html URL. Let me move the check in there, rename it `rewriteHtmlExtensionIfFound(str)`, and then the constants are no longer needed!

src/urls/util.ts
```diff
@@ -53,8 +30,11 @@ function isBowerDependencyUrl(htmlUrl: OriginalDocumentUrl): boolean {
 /**
  * Rewrite a url to replace a `.js` file extension with `.html`.
  */
-function fixHtmlExtension(htmlUrl: string): string {
-  return htmlUrl.substring(0, htmlUrl.length - htmlExtension.length) + '.js';
+function fixHtmlExtensionIfFound(htmlUrl: string): string {
```
rictic: Relatively minor, but IMO a very short function that's called only once is a candidate for inlining. Documenting the behavior and invariants of the function, and using that knowledge when reasoning around the call site is more cost than benefit.

FredKSchott: I agree, there's already a comment at the call site so it's not like there's much of a clarity win either. Moving to inline...

justinfagnani: I disagree a little with this. There's a benefit from abstracting away both the operation and the cast to a new interface. There's also a benefit to abstracting _all_ path/URL conversions, not just the ones that happen more than once.

I think the callsite comment can go away with a good name for the function. "fix" probably isn't the right verb here, since the URL isn't broken, it's just referring to a file that's being converted to JS.

rictic: There's benefit, and there's cost. Moving the code into a function creates an abstraction, and in this case it makes the code harder to read and reason about. The benefit to the abstraction seems nebulous and pings my YAGNI-senses. Maybe we'll want to change and audit all path/url conversions, but probably not.

If we went whole hog and used branded string types for the different kinds of URLs here I think it would tip the scales for me, as then the abstraction would pay us back with type errors and find references and such.

FredKSchott: Given that this was a 1 time, 1-4 line operation, inlining it makes the most sense to me. There's always a time and place for simple operations to get their own helpers, but I'd only explicitly ask to see one for something very tricky or gross or complex (more docs than code?).

Regardless, now this is needed in a second place (see discussion in #219) so it's a bit of a moot point now.

PR #223, opened 2017-11-07
[ProjectConverter] Remove no-longer-working tests (temporarily)

FredKSchott:
*Builds off of #222: "[ProjectConverter] Add dependency map lookup function"*

## Summary:

This PR removes most tests :( They're about to lose compatibility with the conversion interface, but don't worry, they'll be added back at the end. 

[Meta: This is the first PR that would be better of getting merged into `master` along with the future work. Once it's base is merged to master, I'll create a new base to merge the rest of the PRs into, which can then be merged back in to master all-together.]


FredKSchott:
On second thought, this is pretty straight forward, I'll just include this in the big PR.

PR #224, opened 2017-11-07
[ProjectConverter] Refactor BaseConverter into UrlHandler, ProjectConverter

FredKSchott:
*Builds off of #222: "[ProjectConverter] Remove no-longer-working tests (temporarily)"*

## Summary:

This PR is where the main refactoring work occurs! Here are the two new players introduced in this PR:

### UrlHandler

```ts
interface UrlHandlerInterface {
    getPackageNameForUrl(url: OriginalDocumentUrl): string;
    getPackageTypeForUrl(url: OriginalDocumentUrl): PackageType;
    isImportInternal(fromUrl: ConvertedDocumentUrl, toUrl: ConvertedDocumentUrl): boolean;
    getNameImportUrl(url: ConvertedDocumentUrl): ConvertedDocumentUrl;
    getPathImportUrl(fromUrl: ConvertedDocumentUrl, toUrl: ConvertedDocumentUrl): string;
    convertUrl(url: OriginalDocumentUrl): ConvertedDocumentUrl;
}

class PackageUrlHandler implements UrlHandlerInterface { ... }
class WorkspaceUrlHandler implements UrlHandlerInterface { ... }
```

`UrlHandlerInterface` implementations are built for specific project layouts: currently Workspace & Single Package. By understanding the location of files and dependencies of the layout under conversion, UrlHandlers can handle all the layout-specific logic needed for a proper conversion. 

All other pieces of Polymer Modulizer can be layout-agnostic by delegating to a specific UrlHandler.

### ProjectConverter

```ts
class ProjectConverter {
    readonly urlHandler: UrlHandlerInterface;
    readonly conversionSettings: ConversionSettings;

    convertDocument(document: Document): void;
    convertDocumentToJs(document: Document, visited: Set<OriginalDocumentUrl>): void;
    convertDocumentToHtml(document: Document, visited: Set<OriginalDocumentUrl>): void;
    getResults(): Map<ConvertedDocumentFilePath, string | undefined>;
}
```

The spiritual successor to `ConversionMetadata`, refactored to soley provide an interface over higher-level conversions. Kicks off a document conversion and collects results across multiple file/package conversion.


src/document-converter.ts
```diff
@@ -1041,49 +1047,75 @@ export class DocumentConverter {
...
+
+  /**
+   * Converts the URL for a script that is already being loaded in a
+   * pre-conversion HTML document via the <script> tag. This is similar to
+   * convertDocumentUrl(), but can skip some of the more complex .html -> .js
+   * conversion/rewriting.
+   */
+  private convertScriptUrl(oldUrl: OriginalDocumentUrl): ConvertedDocumentUrl {
+    // TODO(fks): This can be removed later if type-checking htmlUrl is enough
+    if (oldUrl.startsWith('.') || oldUrl.startsWith('/')) {
```
usergenic: Same logic as before.  Even if not using a regexp, this check should probably still be methodized with an intention-revealing name.

src/document-converter.ts
```diff
@@ -1041,49 +1046,75 @@ export class DocumentConverter {
...
    * original HTML import URL is given, attempt to match the format of that
    * import URL as much as possible. For example, if the original import URL was
    * an absolute path, return an absolute path as well.
+   *
+   * TODO(fks) 10-26-2017: Make this run on Windows/Non-Unix systems.
    */
   private formatImportUrl(
-      jsRootUrl: ConvertedDocumentUrl, originalHtmlImportUrl?: string) {
-    // Return an absolute URL path if the original HTML import was absolute
+      jsRootUrl: ConvertedDocumentUrl, originalHtmlImportUrl?: string): string {
```
usergenic: `jsRootUrl` ?  Root of what?  What's its role in this function...  more a "to" or "destination" root or something...?

src/document-converter.ts
```diff
@@ -1041,49 +1046,75 @@ export class DocumentConverter {
...
-    // folder by referencing files up an additional level.
-    if (isPackageScoped) {
-      if (importUrl.startsWith('./')) {
-        importUrl = '../' + importUrl.slice('./'.length);
-      } else {
-        importUrl = '../' + importUrl;
-      }
-    }
-    return importUrl;
+    // Otheriwse, return the external import URL formatted for paths.
```
usergenic: `Otherwise`

src/document-converter.ts
```diff
@@ -1137,8 +1168,9 @@ export class DocumentConverter {
       const references = importedReferences.get(jsImplicitImportUrl);
       const namedExports =
           new Set(IterableX.from(references || []).map((ref) => ref.target));
-
       const jsFormattedImportUrl = this.formatImportUrl(jsImplicitImportUrl);
+
+
```
usergenic: why so spacey~??~? :-p

src/project-converter.ts
```diff
@@ -0,0 +1,159 @@
...
+    this.urlHandler = urlHandler;
+    this.conversionSettings = conversionSettings;
+  }
+
+  /**
+   * Convert a document and any of its dependencies. The output format (JS
+   * Module or HTML Document) is determined by whether the file is included in
+   * conversionSettings.includes.
+   */
+  convertDocument(document: Document) {
```
usergenic: Maybe indicate that this `document` variable is always an `HtmlDocument` or `ParsedHtmlDocument` or whatever that class is -- i.e. this is not handed a `JavaScriptDocument` as input right now.

src/project-converter.ts
```diff
@@ -0,0 +1,159 @@
...
+   * Check if a document is explicitly excluded or has already been converted
+   * to decide if it should be converted or skipped.
+   */
+  shouldConvertDocument(document: Document): boolean {
+    const documentUrl = getDocumentUrl(document);
+    if (this.conversionResults.has(documentUrl)) {
+      return false;
+    }
+    for (const exclude of this.conversionSettings.excludes) {
+      if (documentUrl.endsWith(exclude)) {
```
usergenic: Is the format of the `exclude` string such that it is always preceded by a delimiter like '/' such that `endsWith` will reliably match `apples` to `my/apples` but not `crapples`?

src/project-converter.ts
```diff
@@ -0,0 +1,159 @@
...
+    const documentConverter = new DocumentConverter(this, document, visited);
+    const newModule = documentConverter.convertAsToplevelHtmlDocument();
+    this.handleConversionResult(newModule);
+  }
+
+  /**
+   * A private instance method for handling new conversion results, exports,
+   * etc.
+   */
+  private handleConversionResult(newModule: ConversionResult): void {
```
usergenic: `_handleConversionResult`

src/project-converter.ts
```diff
@@ -0,0 +1,159 @@
...
+   * This method collects the results after all documents are converted. It
+   * handles out some broken edge-cases (ex: shadycss) and sets empty map
+   * entries for files to be deleted.
+   */
+  getResults(): Map<ConvertedDocumentFilePath, string|undefined> {
+    const results = new Map<ConvertedDocumentFilePath, string|undefined>();
+
+    for (const convertedModule of this.conversionResults.values()) {
+      if (convertedModule.originalUrl.endsWith(
+              'shadycss/entrypoints/apply-shim.js') ||
```
usergenic: Maybe this can be `/shadycss/entrypoints/apply-shim.js` in case I have an element using `notshadycss`?  Far fetched?

src/urls/package-url-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+
+import {lookupDependencyMapping} from '../manifest-converter';
+
+import {ConvertedDocumentFilePath, ConvertedDocumentUrl, OriginalDocumentUrl, PackageType} from './types';
+import {UrlHandlerInterface} from './url-handler-interface';
+import {getRelativeUrl} from './util';
+
+const isInBowerComponentsRegex = /(\b|\/|\\)(bower_components)(\/|\\)/;
+const isInNodeModulesRegex = /(\b|\/|\\)(node_modules)(\/|\\)/;
```
usergenic: `\b` matches `.` and ` ` and `:` for example.  Not sure if this is what you are really after.  Maybe these should just be `/(^|\/|\\)(bower_components)(\/|\\)/` and `/(^|\/|\\)(node_modules)(\/|\\)/`?

src/urls/package-url-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+   * Get the converted NPM name for a package, given the original document URL
+   * of any file that lives in that package.
+   */
+  getPackageNameForUrl(url: OriginalDocumentUrl) {
+    if (PackageUrlHandler.isUrlInternalToPackage(url)) {
+      return this.packageName;
+    } else {
+      // For a a single package layout, the Bower package name is included in
+      // the URL. ie: bower_components/PACKAGE_NAME/...
+      const basePackageName = url.split('/')[1];
```
usergenic: I feel like TypeScript should have yelled at you for not ensuring that `basePackageName` is defined (i.e. if url didn't contain a `/`) since `lookupDependencyMapping` demands a string for its first argument.

We must have less strict settings or something on tsconfig/tslint for this project?  I'd have expected `url.split('/')[1]!` or something.

src/urls/util.ts
```diff
@@ -15,24 +15,12 @@
...
-  return htmlUrl.startsWith('bower_components/') ||
-      htmlUrl.startsWith('./bower_components/');
-}
-
-/**
  * Rewrite a url to replace a `.html` file extension with `.js`, if found.
  */
-function replaceHtmlExtensionIfFound(url: string): string {
+export function replaceHtmlExtensionIfFound(url: string): string {
   if (url.endsWith('.html')) {
```
usergenic: this whole bit feels like it should have been written as
```js
return url.replace(/\.html$/, '.js');
```

Do we not favor regex for some performance reason?

src/document-converter.ts
```diff
@@ -1041,49 +1047,75 @@ export class DocumentConverter {
   }
 
   /**
+   * Converts an HTML Document's path from old world to new. Use new NPM naming
+   * as needed in the path, and change any .html extension to .js.
+   */
+  private convertDocumentUrl(htmlUrl: OriginalDocumentUrl):
+      ConvertedDocumentUrl {
+    // TODO(fks): This can be removed later if type-checking htmlUrl is enough
+    if (htmlUrl.startsWith('.') || htmlUrl.startsWith('/')) {
```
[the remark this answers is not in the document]
FredKSchott: This check isn't even really right anymore, extracted into a helper and simplified, lmkwyt

src/project-converter.ts
FredKSchott: Some things in flight across multiple PRs should also be changed to `_`-prefix, adding a TODO to go back after everything is merged.

src/project-converter.ts
FredKSchott: in workspace mode their originalUrl is literally `"shadycss/entrypoints/apply-shim.js"`. But if we never write this, maybe we should just add it to excludes? Not sure why this is being handled seperately.

src/urls/util.ts
FredKSchott: This went through a few iterations (see some past PRs) but you're right this is much cleaner. 💯

src/project-converter.ts
FredKSchott: Is there an HtmlDocument type? I don't see one exported by the analyzer, and `analysis.getFeatures({kind: 'html-document'})` is typed to return a Set of `Document`s.

src/urls/package-url-handler.ts
FredKSchott: lets chat in person, want to make sure this supports both URL formats ('./' prefixed & not prefixed)

src/project-converter.ts
FredKSchott: Oh right, we want to preserve any imports of this file (excludes completely removes/ignores them). 

I'm going to comment this and mark as TODO for now, and we can revisit after this PR. Current logic is *okay*, and it's unclear if this list will need to be user-definable at some point or can live as a hardcoded one-off.

src/project-converter.ts
FredKSchott: This is similar to https://github.com/Polymer/polymer-modulizer/pull/224#discussion_r150687180. `endsWith()` is a more flexible matcher (it works on both package and workspace layouts) but I agree that it might be too flexible. Lets make this an exact `===` matcher for now, and I'll add an issue to convert to something thats both explicit & flexible, like glob matching.

src/document-converter.ts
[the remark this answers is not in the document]
usergenic: Win-win!

src/project-converter.ts
usergenic: I just now see that the `Document` model returned in the `getDocument` call of an `Analysis` is basically a wrapper class and that it exposes the typed versions through the `parsedDocument` getter, so you can't express the constraint here through typings AFAICT.  So maybe just a comment, if that.

usergenic:
👍

src/document-converter.ts
```diff
@@ -151,50 +152,50 @@ function getImportDeclarations(
...
-   * Returns the HTML Imports of a document, except imports to documents
-   * specifically excluded in the AnalysisConverter.
+   * Returns the HTML Imports from a document, except imports to documents
+   * specifically excluded in the ConversionSettings.
    *
    * Note: Imports that are not found are not returned by the analyzer.
    */
   private getHtmlImports() {
     return IterableX.from(this.document.getFeatures({kind: 'html-import'}))
         .filter((f: Import) => {
```
usergenic: golf!
```ts
.filter((f: Import) => !this.conversionSettings.excludes.has(f.document.url));
```

src/test/project-converter_test.ts
```diff
@@ -125,20 +146,21 @@ suite('AnalysisConverter', () => {
...
           <script></script>
         `,
         'dep.html': `<h1>Hi</h1>`,
-        'bower_components/dep.html': `<h1>Hi</h1>`,
+        'bower_components/dep/dep.html': `<h1>Hi</h1>`,
       });
       const expectedWarnings = [
-        `WARN: bower->npm mapping for "dep.html" not found`,
-        `WARN: bower->npm mapping for "dep.html" not found`,
+        `WARN: bower->npm mapping for "dep" not found`,
```
usergenic: I thought there was some kind of memoization in the lookup to reduce duplicate warnings?  Maybe I misunderstood.  In any case it might be good to state in a comment why 3 duplicate warnings are expected.

src/urls/package-url-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+        !url.startsWith('./node_modules/');
+  }
+
+  constructor(packageName: string, packageType?: PackageType) {
+    this.packageName = packageName;
+    this.packageType = packageType || 'element';
+  }
+
+  /**
+   * Get the converted NPM name for a package, given the original document URL
```
justinfagnani: mention that this extracts the Bower name, then runs through the dependencyMap?

src/urls/package-url-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+   */
+  static isUrlInternalToPackage(url: ConvertedDocumentUrl|OriginalDocumentUrl|
+                                ConvertedDocumentFilePath) {
+    return !url.startsWith('bower_components/') &&
+        !url.startsWith('./node_modules/');
+  }
+
+  constructor(packageName: string, packageType?: PackageType) {
+    this.packageName = packageName;
+    this.packageType = packageType || 'element';
```
justinfagnani: Add a default value to the parameter instead

src/urls/package-url-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+  readonly packageName: string;
+  readonly packageType: PackageType;
+
+  /**
+   * Helper function to check if a file URL is internal to the main package
+   * being converted (vs. a dependency).
+   */
+  static isUrlInternalToPackage(url: ConvertedDocumentUrl|OriginalDocumentUrl|
+                                ConvertedDocumentFilePath) {
+    return !url.startsWith('bower_components/') &&
```
justinfagnani: Do we know we don't have cases like `'./bower_components'` or `'node_modules'`?

src/urls/package-url-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * Get the converted NPM name for a package, given the original document URL
+   * of any file that lives in that package.
+   */
+  getPackageNameForUrl(url: OriginalDocumentUrl) {
+    // If url contains no directories it must be internal. Explicitly check
+    // `basePackageName` here since it is needed below.
+    const urlParts = url.split('/');
+    const basePackageName = urlParts[1] as (string | undefined);
```
justinfagnani: Won't this fail for URLs with schemes? I think we should use the URL object, though it would require passing a base URL around: https://nodejs.org/api/url.html#url_constructor_new_url_input_base TODO/issue?

justinfagnani: try to avoid out-of-bounds array indexing, check for length

src/urls/package-url-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+    return ('./' + newUrlPieces.join('/')) as ConvertedDocumentUrl;
+  }
+
+  /**
+   * Get the formatted import URL between two ConvertedDocumentUrls.
+   */
+  getPathImportUrl(fromUrl: ConvertedDocumentUrl, toUrl: ConvertedDocumentUrl):
+      string {
+    const isPackageNameScoped = this.packageName.includes('/');
```
justinfagnani: `this.packageName.startsWith('@')`?

src/urls/package-url-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+    if (!depInfo) {
+      return basePackageName;
+    }
+    return depInfo.npm;
+  }
+
+  /**
+   * Get the "type" for the package where a file lives, based on it's URL.
+   */
+  getPackageTypeForUrl(url: OriginalDocumentUrl) {
```
justinfagnani: This isn't really a package type so much as a URL type. When is this method called? If it's per-package rather than per URL, could we get rid of this concept at the package level?

src/document-converter.ts
```diff
@@ -1041,49 +1044,75 @@ export class DocumentConverter {
...
+    return this.urlHandler.convertUrl(oldUrl);
+  }
+
+  /**
    * Format an import from the current document to the given JS URL. If an
    * original HTML import URL is given, attempt to match the format of that
    * import URL as much as possible. For example, if the original import URL was
    * an absolute path, return an absolute path as well.
+   *
+   * TODO(fks) 10-26-2017: Make this run on Windows/Non-Unix systems.
```
justinfagnani: Let's add a Window support issue and reference is here. No need for dates in todos.

src/document-converter.ts
```diff
@@ -1041,49 +1044,75 @@ export class DocumentConverter {
...
-    let importUrl = getRelativeUrl(this.convertedUrl, jsRootUrl);
-    // If this document is an external dependency, or if this document is
-    // importing a local file, just return normal relative URL between the two
-    // files.
-    if (!isImportFromLocalFile || isImportToLocalFile) {
-      return importUrl;
+    // If the import is contained within a single package (internal), return
+    // a path-based import.
+    if (this.urlHandler.isImportInternal(this.convertedUrl, toUrl)) {
+      return this.urlHandler.getPathImportUrl(this.convertedUrl, toUrl);
```
justinfagnani: this is the same as like 1115, why the `if`?

src/project-converter.ts
justinfagnani: You can definitely check that you're handed an HTML document though, I'd do that.

src/project-converter.ts
```diff
@@ -0,0 +1,165 @@
...
+export class ProjectConverter {
+  readonly urlHandler: UrlHandlerInterface;
+  readonly conversionSettings: ConversionSettings;
+
+  /**
+   * A cache of all JS Exports by namespace, to map implicit HTML imports to
+   * explicit named JS imports.
+   */
+  readonly namespacedExports = new Map<string, JsExport>();
+  /**
```
justinfagnani: add newline

src/project-converter.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /**
+   * Convert a document and any of its dependencies. The output format (JS
+   * Module or HTML Document) is determined by whether the file is included in
+   * conversionSettings.includes.
+   */
+  convertDocument(document: Document) {
+    try {
+      this.conversionSettings.includes.has(document.url) ?
```
justinfagnani: did clang do this? I definitely prefer the `?` and `:` to be at the beginning of the lines if not.

src/project-converter.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+
+  /**
+   * Check if a document is explicitly excluded or has already been converted
+   * to decide if it should be converted or skipped.
+   */
+  shouldConvertDocument(document: Document): boolean {
+    const documentUrl = getDocumentUrl(document);
+    if (this.conversionResults.has(documentUrl)) {
```
justinfagnani:
```ts
return !this.conversionResults.has(documentUrl) && !this.conversionSettings.excludes.has(documentUrl);
```

src/project-converter.ts
```diff
@@ -0,0 +1,165 @@
...
+      return false;
+    }
+    if (this.conversionSettings.excludes.has(documentUrl)) {
+      return false;
+    }
+    return true;
+  }
+
+  /**
+   * Specifically convert an HTML document to a JS module. Useful during
```
justinfagnani: Can remove "Specifically"

src/project-converter.ts
```diff
@@ -0,0 +1,165 @@
...
+    if (!this.shouldConvertDocument(document)) {
+      return;
+    }
+    const documentConverter = new DocumentConverter(this, document, visited);
+    const newModule = documentConverter.convertToJsModule();
+    this._handleConversionResult(newModule);
+  }
+
+  /**
+   * Specifically convert an HTML document without changing the file type
```
justinfagnani: Can remove "Specifically"

src/project-converter.ts
```diff
@@ -0,0 +1,165 @@
...
+   * (changes imports and inline scripts to modules as necessary.) Useful during
```
justinfagnani: move `.` outside parens

src/project-converter.ts
```diff
@@ -0,0 +1,165 @@
...
+  getResults(): Map<ConvertedDocumentFilePath, string|undefined> {
+    const results = new Map<ConvertedDocumentFilePath, string|undefined>();
+
+    for (const convertedModule of this.conversionResults.values()) {
+      // These files were already ES6 modules, so don't write our conversions.
+      // Can't be handled in `excludes` because we want to preserve imports of.
+      // TODO(fks): This is hacky, ProjectConverter isn't supposed to know about
+      //  project layout / file location. Move into URLHandler, potentially make
+      //  its own `excludes`-like settings option.
+      if (convertedModule.originalUrl ===
```
justinfagnani: With 4 items, I'd go ahead and make a Set containing these, even if it's going to change in the future. Also, can we just detect that these are modules and not convert them?

src/project-converter.ts
```diff
@@ -0,0 +1,165 @@
...
+              'shadycss/entrypoints/apply-shim.js' ||
+          convertedModule.originalUrl ===
+              'bower_components/shadycss/entrypoints/apply-shim.js' ||
+          convertedModule.originalUrl ===
+              'shadycss/entrypoints/custom-style-interface.js' ||
+          convertedModule.originalUrl ===
+              'bower_components/shadycss/entrypoints/custom-style-interface.js') {
+        continue;
+      }
+      if (convertedModule.keepOriginal !== true) {
```
justinfagnani: why not `=== false` or just `!` if this field is always a boolean? If `undefined` is a valid value and means to keep, then this property should be renamed to `deleteOriginal` where `undefined` means false.

src/document-converter.ts
FredKSchott: ⛳️  nice catch

src/test/project-converter_test.ts
FredKSchott: Haha almost, I was explaining that that would be a nice feature that's easier to implement now :) Still TODO.

src/document-converter.ts
FredKSchott: Yea this is weird, it comes from the addition of getNameImportUrl() in the next PR. Fixing...

src/project-converter.ts
FredKSchott: clang

src/project-converter.ts
FredKSchott: Conversion and analysis of a module (what are its exports, namespaces, etc) are coupled at this point. Can't do one without the other.

src/project-converter.ts
FredKSchott: Yea normally I would just make this property explictly required, but this option only makes sense if the file is not being overwritten.

Will rename to deleteOriginal

src/urls/package-url-handler.ts
FredKSchott: no. Will document

src/urls/package-url-handler.ts
FredKSchott: No, this is only ever called on local documents / OriginalDocumentUrl is only  a local/relative URL.

FredKSchott: Why? I actually did have had a length check, a return statement, and then a later variable assignment, but this is more explicit imo.

src/urls/package-url-handler.ts
FredKSchott: Yes, this is planned but v. v. much out of scope for this PR.

src/urls/package-url-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+      return this.packageType;
+    } else {
+      return 'element';
+    }
+  }
+
+  /**
+   * Check if two URLs are internal within the same package.
+   */
+  isImportInternal(fromUrl: ConvertedDocumentUrl, toUrl: ConvertedDocumentUrl) {
```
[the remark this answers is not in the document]
FredKSchott:
```ts
    if (from.startsWith('./node_modules') && to.startsWith('./node_modules')) {
      return true;
    }
    if (!from.startsWith('./node_modules') && !to.startsWith('./node_modules')) {
      return true;
    }
```

oof I really like `fromUrl` & `toUrl` for readability, `from` and `to` are way to vague. Characters are free

src/urls/package-url-handler.ts
```diff
@@ -0,0 +1,149 @@
...
+    return false;
+  }
+
+  /**
+   * Update a Bower package name in a url (at path index) to its matching npm
+   * package name.
+   */
+  convertUrl(url: OriginalDocumentUrl): ConvertedDocumentUrl {
+    if (PackageUrlHandler.isUrlInternalToPackage(url)) {
+      // TODO(fks): Revisit this format? The analyzer returns URLs without this
```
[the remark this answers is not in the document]
FredKSchott: Yes

src/urls/package-url-handler.ts
justinfagnani: I've been trying to use fast-by-default patterns pervasively recently, mostly to build up the habits so that in situations where it matters we don't have to go back and find and update slow patterns, which may never happen. That's why I'm encouraging explicit checks in conditionals now, rather than relying on truthiness.

In this case, out-of-bounds array access causes a deopt of the array code. Babel recently had some fixes to code like this by the v8 team.

PR #228, opened 2017-11-13
Finish pulling out passes & utility functions from document converter

FredKSchott:
## Summary

There was an original push to organize document conversion "passes" into separate functions that could live outside `document-converter.ts` (already ~1500 lines long). 

- Move the final passes (stateless) off of DocumentConverter prototype and into `passes/*`
- Collect all shared document modifiers & helpers from `util.ts` & `document-converter.ts` into a single place: `document-util.ts`
- Add lots and lots of docs


src/document-converter.ts
```diff
@@ -438,8 +466,8 @@ export class DocumentConverter {
     };
   }
 
-  private rewriteInlineScript(program: estree.Program) {
-    if (this.containsWriteToGlobalSettingsObject(program)) {
+  private rewriteInlineScript(program: Program) {
+    if (containsWriteToGlobalSettingsObject(program)) {
```
justinfagnani: can you add a comment saying _why_ we bail here?

src/document-converter.ts
```diff
@@ -608,73 +604,22 @@ export class DocumentConverter {
...
     }
-    const statements = this.getCodeToInsertDomNodes(genericElements);
+    const statements = createDomNodeInsertStatements(genericElements);
     let insertionPoint = 0;
-    for (const [idx, statement] of enumerate(program.body)) {
-      insertionPoint = idx;
-      if (statement.type === 'ImportDeclaration') {
-        insertionPoint++;  // cover the case where the import is at the end
-        continue;
+    for (let i = 0; i < program.body.length; i++) {
```
justinfagnani: this would be nice as a named function to self-describe what it's doing

src/passes/remove-namespace-initializers.ts
```diff
@@ -14,8 +14,36 @@
 
 import * as astTypes from 'ast-types';
 import * as estree from 'estree';
-import {getMemberOrIdentifierName} from '../util';
+import {getMemberOrIdentifierName} from '../document-util';
 
+/**
+ * Detect certain types of expressions that we deep noops when looking for
```
justinfagnani: "deep noops"? :)

justinfagnani: might be nice to have an example or two

src/passes/remove-namespace-initializers.ts
```diff
@@ -14,8 +14,36 @@
...
+  if (getMemberOrIdentifierName(expression) === assigningTo) {
+    return true;
+  }
+
+  // Most expressions are not noop initializations.
+  return false;
+}
+
+/**
+ * Remove initializers for a set fo namespaces in a program.
```
justinfagnani: fo -> of

src/passes/remove-namespace-initializers.ts
FredKSchott: they're described in the code, but added one to the description for editors

src/document-converter.ts
```diff
@@ -241,18 +275,12 @@ export class DocumentConverter {
...
-    this.rewriteExcludedReferences(program);
-    this.rewriteReferencesToLocalExports(program, exportMigrationRecords);
-    this.rewriteReferencesToNamespaceMembers(
-        program,
-        new Set(IterableX.from(localNamespaceNames)
-                    .concat(
-                        namespaceNames,
-                        )));
+    const allNamespaceNames =
+        new Set([...localNamespaceNames, ...namespaceNames]);
```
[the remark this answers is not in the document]
FredKSchott: I'm okay with this, it allows us to refactor the entire process into a "pass". Definitely a candidate for perf down the road once other lower hanging fruit is gone

src/cli/index.ts
```diff
@@ -61,11 +61,7 @@ const optionDefinitions: commandLineArgs.OptionDefinition[] = [
     defaultValue: 'modulizer_out',
     description: 'The directory to write converted files to.'
   },
-  {
-    name: 'in',
-    type: String,
-    description: 'The directory to convert.'
-  },
+  {name: 'in', type: String, description: 'The directory to convert.'},
```
stramel: Is there a reason you want this arg to be defined on a single line when all the others aren't?

FredKSchott: It's clang auto-formatting, because it can fit on 1 line it gets auto-formatted. You can see for yourself by running `npm run format`

It can be weird (like here, I wouldn't do this change myself, personally) but it prevents us from having to worry/argue about formatting which is a 💯 win.

stramel: Alright, that makes sense! :+1: I forgot to run that command on my PR 🙃

PR #229, opened 2017-11-13
Move "dependencies" responsibilities out of DocumentConverter

FredKSchott:
The lower-level DocumentConverter has a cyclical relationship with the ProjectConverter that runs it. The flow currently looks like this:

1. ProjectConverter spins up a `DocumentConverter` instance to convert a document
2. The first thing `DocumentConverter` does is check its document's imports
3. For each import, `DocumentConverter` calls back up to its stored reference to the `ProjectConverter` that spawned it telling it to convert that dependency document first
4. Above steps are repeated for that dependency, and its dependencies, and its dependencies, etc etc 

Instead of juggling dependency responsibilities across two different objects, `ProjectConverter` should take over the entire dependency handling process.`DocumentConverter` can then be decoupled from any higher-level, multi-document concerns. Instead of having it do its own private lookups / spawning other `DocumentConverter`s of its own, all necessary data is passed to the `DocumentConverter` constructor. Not only is this an easier structure to understand, but it makes `DocumentConverter` much easier to test.


src/project-converter.ts
```diff
@@ -74,7 +74,7 @@ export class ProjectConverter {
    * Check if a document is explicitly excluded or has already been converted
    * to decide if it should be converted or skipped.
    */
-  shouldConvertDocument(document: Document): boolean {
+  private shouldConvertDocument(document: Document): boolean {
```
usergenic: `_shouldConvertDocument`

src/project-converter.ts
```diff
@@ -88,14 +88,45 @@ export class ProjectConverter {
   }
 
   /**
+   * Convert document dependencies. Should be called early in the conversion
+   * process so that it can read this document's dependencies' exports.
+   */
+  private convertDependencies(
```
usergenic: `_convertDependencies`

src/document-converter.ts
```diff
@@ -159,30 +159,34 @@ function getImportDeclarations(
 export class DocumentConverter {
   private readonly originalUrl: OriginalDocumentUrl;
```
usergenic: private properties and methods by convention are prefixed with `_` and this is actually linted for in some of our other projects...  Maybe separate PR but `originalUrl` -> `_originalUrl` etc.

usergenic:
Seems good.  I think my only comments were about private property/method names, but those don't necessarily have to be addressed in this PR since they're a global issue in modulizer right now.

src/document-converter.ts
FredKSchott: Agreed, I'll tackle in a separate PR since there are a few different PRs in flight that this would affect.

src/project-converter.ts
```diff
@@ -87,21 +87,52 @@ export class ProjectConverter {
...
 
+  /**
+   * Convert document dependencies. Should be called early in the conversion
+   * process so that it can read this document's dependencies' exports.
+   */
+  private _convertDependencies(
+      document: Document, visited: Set<OriginalDocumentUrl>) {
+    for (const htmlImport of DocumentConverter.getAllHtmlImports(document)) {
+      // Only convert what we need, ignore excluded/already converted documents.
+      if (!this._shouldConvertDocument(htmlImport.document)) {
```
rictic: Can we remove this check? It looks like `this.convertDocumentToJs` also checks if it should convert, and early exits if not

FredKSchott: We get bad "Cycle in dependency graph" warnings below if this is removed. I'll add that info to the comment to make this more clear

PR #233, opened 2017-11-14
Fix: out directory default to cwd

stramel:
Fixes #140 

Defaults the `outDir` to the cwd instead of `modulizer_out`

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ add the new ones in the current directory:
 
 ```sh
 bower cache clean && bower install
-modulizer --out .
+modulizer
 ```
 
 ### Workspace mode
@@ -171,4 +171,4 @@ cd ../polymer
 modulizer
 ```
 
-The converted files are now in the directory `js_out`
+The converted files are now in the current working directory
--- a/src/cli/command-package.ts
+++ b/src/cli/command-package.ts
@@ -22,7 +22,7 @@ import {exec, logStep} from '../util';
 
 export default async function run(options: CliOptions) {
   const inDir = path.resolve(options.in || process.cwd());
-  const outDir = path.resolve(options.out);
+  const outDir = path.resolve(options.out || process.cwd());
 
   // Ok, we're updating a package in a directory not under our control.
   // We need to be sure it's safe.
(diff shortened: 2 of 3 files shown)


FredKSchott:
We need to figure out which is the "recommended"/default behavior: to overwrite the current directory or write out to a sub directory.

stramel:
Okay, just trying to hit any low hanging fruit and it seemed like current directory was the way y'all were leaning. 

Also, @FredKSchott, I'm noticing that with `modulizer_out` as the `out` with the current state of the code it is creating nested subfolders.

**Expected:**

```
/
|-- modulizer_out/
    |-- demo/
        |-- index.html
    |-- test/
        |-- index.html
        |-- marked-element.html
    |-- index.html
    |-- marked-element.js
    |-- marked-import.js
    |-- package.json
```

**Actual:**

```
/
|-- modulizer_out/
    |-- demo/
         |-- index.html
    |-- modulizer_out/
        |-- demo/
            |-- index.html
        |-- modulizer_out/
            |-- demo/
                |-- index.html
            |-- test/
                |-- index.html
                |-- marked-element.html
            |-- index.html
        |-- test/
            |-- index.html
            |-- marked-element.html
    |-- test/
        |-- index.html
        |-- marked-element.html
    |-- index.html
    |-- marked-element.js
    |-- marked-import.js
    |-- package.json
```

FredKSchott:
Yea I think that's right for now. Can you fix your README change to use that instead of js_out?

stramel:
@FredKSchott Is there anything else you would like addressed here?

stramel:
Ping @fredkschott

justinfagnani:
@FredKSchott and @stramel I kind of think we should just remove the `--out` flag and only support cwd, with the assumption, and maybe a check, that we're in a git repo.

stramel:
@fredkschott I know you had some opinions on not removing the out directory in another issue/PR a while back. Personally, I'm good with whichever.

justinfagnani:
Looking back at this again, I think we should remove package mode for now to reduce the overhead of two modes, then look at bringing it back as options to workspace mode. That would mean closing this PR.

@FredKSchott wdyt?

stramel:
Going to close this. Seems like it may not be the approach you guys wanted and is easy enough to reproduce.

PR #234, opened 2017-11-14
New: Check for SPDX license

### LINKED ISSUE #185: Generate npm-compatible licenses in package.json
Sounds like the direct translation from bower.json is causing warnings in npm: https://github.com/PolymerElements/marked-element/pull/77

stramel:
Fixes #185 

Handles the polymer LICENSE URL => BSD-3-Clause
Warns user if the current license in bower is not a valid SPDX license. Perhaps further improvement could suggest the closest license?

**Open Questions:**

- How should we handle an existing `package.json` and `bower.json`, both having different licenses specified?
- Should we warn if no license has been specified or default to something like `Unlicense`?
- Would it be better to have a tedium process change the license on all PolymerElements from the Polymer License Url to BSD-3-Clause instead of having a custom logic branch built in here?

### DIFF
--- a/custom_typings/spdx-license-list.d.ts
+++ b/custom_typings/spdx-license-list.d.ts
@@ -1,0 +1,4 @@
+declare module "spdx-license-list/simple" {
+  const simple: Set<String>;
+  export = simple;
+}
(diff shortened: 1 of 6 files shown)


justinfagnani:
I think this is a good improvement for now. I think we should have some sort of option to control package.json options better. @FredKSchott is working on manifest merging, so maybe we just prefer the package.json licence field if it exists.

PR #238, opened 2017-11-15
Yarn instead of npm

### LINKED ISSUE #237: Incompatibility with npm
I'm neither able to install nor build modulizer using npm.
Though it totally works with yarn.

web-devel:
Fixes #237

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ window.customElements.define(MyElement.is, MyElement);
 Install polymer-modulizer from npm:
 
 ```
-npm install -g polymer-modulizer
+yarn install -g polymer-modulizer
 ```
 
 `polymer-modulizer` has two modes: package mode, which converts the current directory as a package,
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -13,17 +13,17 @@
   },
   "scripts": {
     "build": "tsc",
-    "test": "npm run build && mocha && npm run lint",
+    "test": "yarn run build && mocha && yarn run lint",
```
stramel: you can do `yarn build && mocha && yarn lint`. `npm` requires the use of `run` keyword before the script name.

README.md
```diff
@@ -66,7 +66,7 @@ window.customElements.define(MyElement.is, MyElement);
 Install polymer-modulizer from npm:
 
 ```
-npm install -g polymer-modulizer
+yarn install -g polymer-modulizer
```
stramel: I think this is also supposed to be `yarn global add polymer-modulizer`

justinfagnani:
We're not requiring Yarn to install modulizer, so any fix to #237 should actually fix whatever problem there is with that particular npm installation.

FredKSchott:
Yea, I'd go a step further and say these should remain `npm run`. 

While we recommend yarn for a lot of things, we still also support `npm`. `npm run` is guaranteed to work for anyone (it comes with Node) while `yarn run` would fail without yarn.

Lets fix the npm issue itself instead!

web-devel:
Thanks for the clarification, it make sense. I think it probably might be better to change `yarn install` →`npm install` in readme then.

PR #209, opened 2017-10-09
Add ability to push to github & npm after modulization is complete

FredKSchott:
This completes the "workspaces" flow for pushing & publishing changes to multiple repos after conversion. This previously had to be done by hand using some janky scripts & bash magic.

[demo!](https://gfycat.com/gifs/detail/MagnificentFreshHartebeest)

Travis won't pass until the two workspaces dependencies are pushed and published:
- [X] https://github.com/Polymer/polymer-workspaces/pull/7
- [X] https://github.com/Polymer/polymer-workspaces/pull/8
- [X] https://github.com/Polymer/polymer-workspaces/pull/10 (important bug fix)

PR #247, opened 2017-12-08
run tests across workspace repos!

FredKSchott:
- `src/test-workspace`: Adds the ability to run tests & `npm install` across repos in a workspace. Packages under conversion are tested against each other (if any are dependencies).
- `src/cli/*`: Presents a menu of actions for the user to do after workspace conversion (right now just "test" & "test (install only)"). Work from #209 to add "push to github" and "publish to npm" will be added here.

screen shot 2017-12-08 at 1 39 43 pm

- Based on https://github.com/Polymer/polymer-workspaces/pull/14, needs to be released to npm before tests will pass. But you can use locally by `npm link`ing that PR.
- Few element tests are passing, but all (or almost all) `npm installs` are working!

### DIFF
--- a/src/cli/command-package.ts
+++ b/src/cli/command-package.ts
@@ -84,7 +84,7 @@ export default async function run(options: CliOptions) {
         }]))['npm-version'] as string;
   }
 
-  logStep('[1/2]', '🌀', `Converting Package...`);
+  logStep(1, 2, '🌀', `Converting Package...`);
   console.log(`Out directory: ${outDir}`);
   await convertPackage({
     inDir: inDir,
@@ -97,5 +97,5 @@ export default async function run(options: CliOptions) {
     cleanOutDir: options.clean!!,
   });
 
-  logStep('[2/2]', '🎉', `Conversion Complete!`);
+  logStep(2, 2, '🎉', `Conversion Complete!`);
 }
--- a/src/cli/command-package.ts
+++ b/src/cli/command-package.ts
@@ -11,7 +11,6 @@
  * subject to an additional IP rights grant found at
  * http://polymer.github.io/PATENTS.txt
  */
-import chalk from 'chalk';
 import * as inquirer from 'inquirer';
 import * as path from 'path';
 import * as semver from 'semver';
@@ -19,7 +18,7 @@ import * as semver from 'semver';
 import {CliOptions} from '../cli';
 import convertPackage from '../convert-package';
 import {readJson} from '../manifest-converter';
-import {exec} from '../util';
+import {exec, logStep} from '../util';
 
 export default async function run(options: CliOptions) {
   const inDir = path.resolve(options.in || process.cwd());
@@ -85,10 +84,8 @@ export default async function run(options: CliOptions) {
         }]))['npm-version'] as string;
   }
 
-  console.log(
-      chalk.dim('[1/2]') + ' 🌀  ' + chalk.magenta(`Converting Package...`));
+  logStep('[1/2]', '🌀', `Converting Package...`);
   console.log(`Out directory: ${outDir}`);
-
   await convertPackage({
     inDir: inDir,
     outDir: outDir,
@@ -100,6 +97,5 @@ export default async function run(options: CliOptions) {
     cleanOutDir: options.clean!!,
   });
 
-  console.log(
-      chalk.dim('[2/2]') + ' 🎉  ' + chalk.magenta(`Conversion Complete!`));
+  logStep('[2/2]', '🎉', `Conversion Complete!`);
 }
--- a/src/cli/command-workspace.ts
+++ b/src/cli/command-workspace.ts
@@ -12,13 +12,15 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import chalk from 'chalk';
 import * as fs from 'fs';
+import * as inquirer from 'inquirer';
 import * as path from 'path';
 import {Workspace} from 'polymer-workspaces';
 
 import {CliOptions} from '../cli';
 import convertWorkspace from '../convert-workspace';
+import {testWorkspace, testWorkspaceInstallOnly} from '../test-workspace';
+import {logStep} from '../util';
 
 const githubTokenMessage = `
 You need to create a github token and place it in a file named 'github-token'.
@@ -31,6 +33,31 @@ Then:
 echo 'PASTE TOKEN HEX HERE' > ./github-token
 `;
 
+/**
+ * Post-Conversion steps that the user can select to run after workspace
+ * conversion.
+ */
+enum PostConversionStep {
+  Test = 'test',
+  TestInstallOnly = 'test (install only)',
+  Quit = 'quit',
+}
+
+/**
+ * Create an array of post-conversion steps to run automatically from the given
+ * CLI options. For example, when `--test` is provided the "test"
+ * post-conversion step should be run without prompting.
+ *
+ * Steps should be run in the order returned.
+ */
+function postConversionStepsFromCliOptions(options: CliOptions):
+    PostConversionStep[] {
+  const steps = [];
+  if (options.test === true) {
+    steps.push(PostConversionStep.Test);
+  }
+  return steps;
+}
 
 /**
  * Checks for github-token in the RunnerOptions and if not specified, will look
@@ -56,9 +83,7 @@ function loadGitHubToken(): string|null {
 
 export default async function run(options: CliOptions) {
   const workspaceDir = path.resolve(options['workspace-dir']);
-  console.log(
-      chalk.dim('[1/3]') + ' 🚧  ' +
-      chalk.magenta(`Setting Up Workspace "${workspaceDir}"...`));
+  logStep('[1/3]', '🚧', `Setting Up Workspace "${workspaceDir}"...`);
 
   if (!options['npm-version']) {
     throw new Error('--npm-version required');
@@ -76,7 +101,7 @@ export default async function run(options: CliOptions) {
     dir: workspaceDir,
   });
 
-  const reposToConvert = await workspace.init(
+  const {workspaceRepos: reposToConvert} = await workspace.init(
       {
         include: options['repo']!,
         exclude: options['exclude'],
@@ -86,17 +111,59 @@ export default async function run(options: CliOptions) {
         verbose: true,
       });
 
-  console.log(
-      chalk.dim('[2/3]') + ' 🌀  ' +
-      chalk.magenta(`Converting ${reposToConvert.length} Package(s)...`));
+  await workspace.installBowerDependencies();
 
-  await convertWorkspace({
+  logStep('[2/3]', '🌀', `Converting ${reposToConvert.length} Package(s)...`);
+
+  const convertedPackages = await convertWorkspace({
     workspaceDir,
     npmImportStyle: options['import-style'],
     packageVersion: npmPackageVersion,
     reposToConvert,
   });
 
-  console.log(
-      chalk.dim('[3/3]') + ' 🎉  ' + chalk.magenta(`Conversion Complete!`));
+  logStep('[3/3]', '🎉', `Conversion Complete!`);
+
+  // Loop indefinitely here so that we can control the function exit via the
+  // user prompt.
+  const todoConversionSteps = postConversionStepsFromCliOptions(options);
+  while (true) {
+    // Pull off a "to-do" post-conversion step if any were provided from the
+    // command line, otherwise prompt the user for one.
+    const stepSelection = todoConversionSteps.shift() ||
+        (await inquirer.prompt([{
+          type: 'list',
+          name: 'post-conversion-step',
+          message: ' What do you want to do now?',
+          choices: [
+            PostConversionStep.Test,
+            PostConversionStep.TestInstallOnly,
+            PostConversionStep.Quit
+          ],
+        }]))['post-conversion-step'] as string;
+    switch (stepSelection) {
+      case PostConversionStep.Test:
+        await testWorkspace(convertedPackages, {
+          workspaceDir,
+          packageVersion: npmPackageVersion,
+          reposToConvert,
+        });
+        break;
+      case PostConversionStep.TestInstallOnly:
+        await testWorkspaceInstallOnly(convertedPackages, {
+          workspaceDir,
+          packageVersion: npmPackageVersion,
+          reposToConvert,
+        });
+        break;
+      // TODO(fks): Add 'push to github' support
+      // TODO(fks): Add 'publish to npm' support
+      case PostConversionStep.Quit:
+        console.log('👋  Goodbye.');
+        return;
+      default:
+        console.log(`ERR: option "${stepSelection}" not recognized`);
+        break;
+    }
+  }
 }
--- a/src/cli/command-workspace.ts
+++ b/src/cli/command-workspace.ts
@@ -102,17 +102,13 @@ export default async function run(options: CliOptions) {
   const workspace = new Workspace({
     token: githubToken,
     dir: workspaceDir,
+    match: options['repo']!,
+    exclude: options['exclude'],
+    fresh: options['clean'],
+    verbose: true,
   });
 
-  const {workspaceRepos: reposToConvert} = await workspace.init(
-      {
-        include: options['repo']!,
-        exclude: options['exclude'],
-      },
-      {
-        fresh: options['clean'],
-        verbose: true,
-      });
+  const {workspaceRepos: reposToConvert} = await workspace.init();
 
   await workspace.installBowerDependencies();
 
--- a/src/cli/command-workspace.ts
+++ b/src/cli/command-workspace.ts
@@ -38,9 +38,9 @@ echo 'PASTE TOKEN HEX HERE' > ./github-token
  * conversion.
  */
 enum PostConversionStep {
-  Test = 'test',
-  TestInstallOnly = 'test (install only)',
-  Quit = 'quit',
+  Test = 'Install dependencies and run tests',
+  TestInstallOnly = 'Install dependencies only',
+  Exit = 'Exit',
 }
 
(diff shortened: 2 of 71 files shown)


src/convert-workspace.ts
```diff
@@ -91,48 +89,68 @@ function configureAnalyzer(options: WorkspaceConversionSettings) {
...
-      console.error(err);
-    }
-  }
+  // Commit all changes to a staging branch for easy state resetting.
+  // Useful when performing actions that modify the repo, like installing deps.
+  const commitResults = await run(options.reposToConvert, async (repo) => {
+    await repo.git.createBranch('polymer-modulizer-staging');
+    await exec(repo.dir, 'git', ['add', '-A']);
+    // TODO(fks): Add node_modules to .gitignore, if not found
+    await exec(repo.dir, 'echo', ['reset', '--', 'node_modules/']);
```
stramel: Is this supposed to be `git` instead of `echo`?

src/convert-workspace.ts
```diff
@@ -91,48 +89,68 @@ function configureAnalyzer(options: WorkspaceConversionSettings) {
...
-    try {
-      await writeNpmSymlink(options, repo);
-    } catch (err) {
-      console.log(`Error in npm symlink creation:`);
-      console.error(err);
-    }
-  }
+  // Commit all changes to a staging branch for easy state resetting.
+  // Useful when performing actions that modify the repo, like installing deps.
+  const commitResults = await run(options.reposToConvert, async (repo) => {
```
stramel: I'm assuming we aren't doing this and the post coversion stuff on packages just workspaces?

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+    console.log(chalk.dim(`yarn cache clean: ${stderr}`));
+  }
+
+  const batchResults = await run(reposUnderTest, async (repo) => {
+    await exec(repo.dir, 'git', ['checkout', '-B', localDependenciesBranch]);
+    writeTestingPackageJson(repo, localConversionMap, options.packageVersion);
+    await exec(
+        repo.dir,
+        'git',
+        ['commit', '-a', '-m', 'testing commit', '--allow-empty']);
```
stramel: You could combine `'-a', '-m'` to `'-am'` if you wanted.

src/cli/command-package.ts
```diff
@@ -85,10 +84,8 @@ export default async function run(options: CliOptions) {
         }]))['npm-version'] as string;
   }
 
-  console.log(
-      chalk.dim('[1/2]') + ' 🌀  ' + chalk.magenta(`Converting Package...`));
+  logStep('[1/2]', '🌀', `Converting Package...`);
```
justinfagnani: Eventually it might be nice to use (or build a simple one) a task runner API that does this.

src/cli/command-workspace.ts
```diff
@@ -31,6 +33,31 @@ Then:
 echo 'PASTE TOKEN HEX HERE' > ./github-token
 `;
 
+/**
+ * Post-Conversion steps that the user can select to run after workspace
+ * conversion.
+ */
+enum PostConversionStep {
+  Test = 'test',
```
justinfagnani: I'd suggest slightly different prompts:

- Install dependencies and run tests
- Install dependencies only
- Exit

src/cli/command-workspace.ts
```diff
@@ -31,6 +33,31 @@ Then:
...
+
+/**
+ * Create an array of post-conversion steps to run automatically from the given
+ * CLI options. For example, when `--test` is provided the "test"
+ * post-conversion step should be run without prompting.
+ *
+ * Steps should be run in the order returned.
+ */
+function postConversionStepsFromCliOptions(options: CliOptions):
+    PostConversionStep[] {
```
justinfagnani: what about an `--install` flag?

src/cli/command-workspace.ts
```diff
@@ -86,17 +111,59 @@ export default async function run(options: CliOptions) {
...
+  // user prompt.
+  const todoConversionSteps = postConversionStepsFromCliOptions(options);
+  while (true) {
+    // Pull off a "to-do" post-conversion step if any were provided from the
+    // command line, otherwise prompt the user for one.
+    const stepSelection = todoConversionSteps.shift() ||
+        (await inquirer.prompt([{
+          type: 'list',
+          name: 'post-conversion-step',
+          message: ' What do you want to do now?',
```
justinfagnani: Leading space in message?

src/cli/command-workspace.ts
```diff
@@ -86,17 +111,59 @@ export default async function run(options: CliOptions) {
...
+          choices: [
```
justinfagnani: Does an enum not have a way to get a list of its values?

src/cli/command-workspace.ts
```diff
@@ -86,17 +111,59 @@ export default async function run(options: CliOptions) {
...
+        });
+        break;
+      case PostConversionStep.TestInstallOnly:
+        await testWorkspaceInstallOnly(convertedPackages, {
+          workspaceDir,
+          packageVersion: npmPackageVersion,
+          reposToConvert,
+        });
+        break;
+      // TODO(fks): Add 'push to github' support
```
justinfagnani: file and reference issues

src/convert-workspace.ts
```diff
@@ -91,48 +89,68 @@ function configureAnalyzer(options: WorkspaceConversionSettings) {
...
   // For each repo, convert the relevant HTML documents:
   for (const repo of options.reposToConvert) {
-    const repoDirName = repo.dir.split('/').pop()!;
-    const repoDocuments = htmlDocuments.filter((d) => {
-      return d.url.startsWith(repoDirName) &&
-          !conversionSettings.excludes.has(d.url);
-    });
-    for (const document of repoDocuments) {
+    const repoDirName = path.basename(repo.dir);
+    const bowerPath = path.join(repo.dir, 'bower.json');
```
justinfagnani: `bowerConfigPath`? - to clearly differentiate from the bower_components folder

src/convert-workspace.ts
```diff
@@ -91,48 +89,68 @@ function configureAnalyzer(options: WorkspaceConversionSettings) {
...
+    const packageName = lookupNpmPackageName(bowerPath);
```
justinfagnani: `npmPackageName`?

src/convert-workspace.ts
```diff
@@ -91,48 +89,68 @@ function configureAnalyzer(options: WorkspaceConversionSettings) {
...
+    if (!packageName) {
+      continue;
```
justinfagnani: Log a warning?

src/convert-workspace.ts
```diff
@@ -91,48 +89,68 @@ function configureAnalyzer(options: WorkspaceConversionSettings) {
...
+    }
+    for (const document of htmlDocuments) {
+      if (!document.url.startsWith(repoDirName) ||
```
justinfagnani: You shouldn't need this if you add `externalPackages: false` to the `getFeatures` call.

src/convert-workspace.ts
```diff
@@ -91,48 +89,68 @@ function configureAnalyzer(options: WorkspaceConversionSettings) {
...
-      console.log(`Error in npm symlink creation:`);
-      console.error(err);
-    }
-  }
+  // Commit all changes to a staging branch for easy state resetting.
+  // Useful when performing actions that modify the repo, like installing deps.
+  const commitResults = await run(options.reposToConvert, async (repo) => {
+    await repo.git.createBranch('polymer-modulizer-staging');
+    await exec(repo.dir, 'git', ['add', '-A']);
+    // TODO(fks): Add node_modules to .gitignore, if not found
```
justinfagnani: file and reference issue

src/convert-workspace.ts
```diff
@@ -91,48 +89,68 @@ function configureAnalyzer(options: WorkspaceConversionSettings) {
...
+    await exec(repo.dir, 'echo', ['reset', '--', 'node_modules/']);
+    await repo.git.commit('auto-converted by polymer-modulizer');
+  });
+  commitResults.failures.forEach(logRepoError);
+
+  const symlinkResults = await run(options.reposToConvert, async (repo) => {
```
justinfagnani: Add a comment, maybe update the `writeNpmSymlink` to be more self-describing

src/convert-workspace.ts
```diff
@@ -91,48 +89,68 @@ function configureAnalyzer(options: WorkspaceConversionSettings) {
...
-      await writeNpmSymlink(options, repo);
-    } catch (err) {
-      console.log(`Error in npm symlink creation:`);
-      console.error(err);
-    }
-  }
+  // Commit all changes to a staging branch for easy state resetting.
+  // Useful when performing actions that modify the repo, like installing deps.
+  const commitResults = await run(options.reposToConvert, async (repo) => {
+    await repo.git.createBranch('polymer-modulizer-staging');
```
justinfagnani: Use a constant instead of the literal `'polymer-modulizer-staging'`

src/manifest-converter.ts
```diff
@@ -66,7 +80,10 @@ export function writeJson(json: any, ...pathPieces: string[]) {
  * Given a bower.json manifest, generate a package.json manifest for npm.
  */
 export function generatePackageJson(
-    bowerJson: any, npmName: string, npmVersion: string) {
+    bowerJson: any,
+    npmName: string,
+    npmVersion: string,
+    preferLocal?: Map<string, string>) {
```
justinfagnani: `preferLocal` -> `useLocal`

Can you describe the parameter? What's it a map of?

src/manifest-converter.ts
```diff
@@ -80,25 +97,49 @@ export function generatePackageJson(
     license: bowerJson.license,
     homepage: bowerJson.homepage,
     dependencies: <any>{},
-    devDependencies: <any>{}
+    devDependencies: <any>{},
+    // resolutions needed for wct-browser-legacy
+    resolutions: {
```
justinfagnani: Can this go in a configuration somewhere? Could still be in code, but in a more top-level location?

Also, this is unfortunate... will users get a warning if we're overriding their versions on a conflict?

src/manifest-converter.ts
```diff
@@ -80,25 +97,49 @@ export function generatePackageJson(
...
+    console.warn(
+        `"${bowerJson.license}" is not a valid SPDX license. ` +
+        `You can find a list of valid licenses at https://spdx.org/licenses/`);
   }
 
   for (const bowerPackageName in bowerJson.dependencies) {
     const depInfo = lookupDependencyMapping(bowerPackageName);
-    if (depInfo) {
+    if (!depInfo) {
+      continue;
```
justinfagnani: warning?

src/manifest-converter.ts
```diff
@@ -80,25 +97,49 @@ export function generatePackageJson(
...
       packageJson.dependencies[depInfo.npm] = depInfo.semver;
     }
   }
 
-  // TODO(fks) 07-18-2017: handle devDependencies. Right now wct creates a too
-  // complicated flat dependency install.
+  for (const bowerPackageName in bowerJson.devDependencies) {
+    const depInfo = lookupDependencyMapping(bowerPackageName);
+    if (!depInfo) {
+      continue;
```
justinfagnani: warning?

src/manifest-converter.ts
```diff
@@ -80,25 +97,49 @@ export function generatePackageJson(
...
+          getLocalDependencyValue(preferLocal.get(depInfo.npm)!);
+    } else {
       packageJson.dependencies[depInfo.npm] = depInfo.semver;
     }
   }
 
-  // TODO(fks) 07-18-2017: handle devDependencies. Right now wct creates a too
-  // complicated flat dependency install.
+  for (const bowerPackageName in bowerJson.devDependencies) {
+    const depInfo = lookupDependencyMapping(bowerPackageName);
```
justinfagnani: this is identical code in the two loops, except the field to write to

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+ */
+interface WorkspaceTestSettings extends WorkspaceConversionSettings {}
+
+/**
+ * Setup the workspace repos for testing. Be sure to call restoreRepos() after
+ * testing is complete.
+ */
+async function setupRepos(
+    reposUnderTest: WorkspaceRepo[],
+    localConversionMap: Map<string, string>,
```
justinfagnani: You have a typedef for ConversionResult early in the PR. Reuse that?

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  const batchResults = await run(reposUnderTest, async (repo) => {
+    await exec(repo.dir, 'git', ['checkout', '-B', localDependenciesBranch]);
+    writeTestingPackageJson(repo, localConversionMap, options.packageVersion);
+    await exec(
+        repo.dir,
+        'git',
+        ['commit', '-a', '-m', 'testing commit', '--allow-empty']);
+  }, {concurrency: 10});
+  batchResults.failures.forEach(logRepoError);
```
justinfagnani: We should consider a flag for whether to exit on failures, and what the default should be

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  return [...batchResults.successes.keys()];
```
justinfagnani: Why not return the whole result and let the caller ignore failures if they want to. Otherwise there's no way for the caller to know if there were failures.

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/**
+ * Run `yarn install` to install dependencies in a repo. Note that this creates
+ * a node_modules/ folder & an associated yarn.lock file as side effects.
+ */
+async function installDependencies(reposUnderTest: WorkspaceRepo[]) {
```
justinfagnani: Maybe add `Yarn` to the name here. We'll probably need to add npm support at some point.

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+  const batchResults = await run(reposUnderTest, async (repo) => {
+    const repoDirName = path.basename(repo.dir);
+    const {stderr} = await exec(repo.dir, 'npm', ['install']);
+    if (stderr.length > 0) {
+      console.log(chalk.dim(`${repoDirName}: ${stderr}`));
+    }
+  }, {concurrency: 1});
+
+  batchResults.failures.forEach(logRepoError);
+  return [...batchResults.successes.keys()];
```
justinfagnani: I'd return the full results here too. Basically I'd treat failures a bit like exceptions, in that they should propagate up the call stack, but not actually stop execution. It may be good to add a utility method to chain failures into a new result to help with that.

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/**
+ * Run `wct --npm` in a repo.
+ */
+async function testRepos(reposUnderTest: WorkspaceRepo[]) {
+  const batchResults = await run(reposUnderTest, async (repo) => {
+    const repoDirName = path.basename(repo.dir);
+    const {stdout, stderr} = await exec(repo.dir, 'wct', ['--npm']);
```
justinfagnani: can we stream to the console instead?

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+    console.log(stdout, stderr);
+    if (stdout.length > 0) {
+      console.log(chalk.dim(`${repoDirName}: ${stdout}`));
```
justinfagnani: is this double logging?

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+  return [...batchResults.successes.keys()];
+}
+
+/**
+ * Restore the repos to their proper state.
+ */
+async function restoreRepos(reposUnderTest: WorkspaceRepo[]) {
+  const batchResults = await run(reposUnderTest, async (repo) => {
+    await repo.git.destroyAllUncommittedChangesAndFiles();
+    await repo.git.checkout('polymer-modulizer-staging');
```
justinfagnani: use constant instead of `'polymer-modulizer-staging'`

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+  }, {concurrency: 10});
+
+  batchResults.failures.forEach(logRepoError);
+  return [...batchResults.successes.keys()];
+}
+
+/**
+ * For a given repo, generate a new package.json and write it to disk.
```
justinfagnani: If this is a testing specific package.json, describe how it's different.

src/test-workspace.ts
```diff
@@ -0,0 +1,162 @@
...
+  const testResults = await testRepos(installSuccesses);
+
+  logStep('[4/5]', '🔧', `Restoring Repos...`);
+  await restoreRepos(allRepos);
+
+  logStep('[5/5]', '🔧', `Tests Complete!`);
+  return testResults;
+}
+
+export async function testWorkspaceInstallOnly(
```
justinfagnani: can this be called from `testWorkspace`

src/util.ts
```diff
@@ -77,3 +79,18 @@ export async function exec(
...
+ */
+export function logRepoError(err: Error, repo: WorkspaceRepo) {
+  const repoDirName = path.basename(repo.dir);
+  console.error(`${repoDirName}: ${err.message}`);
+}
+
+/**
+ * Log a user-facing message about progress through some set of steps.
+ */
+export function logStep(stepNum: string, emoji: string, msg: string) {
```
justinfagnani: every call is passing a string like `[1/2]`. I'd take the step numbers as arguments. It'll let you change formatting in one place, or start composing tasks by passing start and total step numbers around.

FredKSchott:
Tests should be passing now on Travis, will wait for the ✅ and then merge

PR #252, opened 2017-12-19
Tag element templates with Polymer.html

### LINKED ISSUE #239: Update template conversion to use Polymer.html template tag
Assuming this proposal goes through: https://github.com/Polymer/polymer/issues/4937

We'll want to output:

```js
import {html, Element as PolymerElement} from '@polymer/polymer/polymer.js';

class MyElement extends PolymerElement {
  template = html`...`;
}
```

justinfagnani:
Fixes #239

### DIFF
--- a/fixtures/packages/iron-icon/expected/iron-icon.js
+++ b/fixtures/packages/iron-icon/expected/iron-icon.js
@@ -1,4 +1,4 @@
-import { Base } from '../../@polymer/polymer/polymer.js';
+import { html, Base } from '../../@polymer/polymer/polymer.js';
 import '../../@polymer/iron-meta/iron-meta.js';
 import '../../@polymer/iron-flex-layout/iron-flex-layout.js';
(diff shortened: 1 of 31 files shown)


justinfagnani:
Er, don't review yet... I forgot some things

justinfagnani:
Ok, updated some unit tests to make sure that the added `Polymer.html` references were indeed being converted properly to JS imports. All seems well there.

I still haven't regenerated the fixtures due to what looks like bad fixture generation with my repo. @FredKSchott could you try to see what happens with fixture generation with a clean checkout on your end?

FredKSchott:
LMK when this is ready for another review

justinfagnani:
@FredKSchott rebase, removed workarounds for not having html-tag.html, and ready for review again.

justinfagnani:
I was thinking of that yesterday, but had already generated these commits and the change was relatively small. Good idea though.

PR #253, opened 2017-12-19
Regenerate fixtures

justinfagnani:


justinfagnani:
@FredKSchott I updated the update-fixtures.ts to override Polymer in the 2nd commit, and regenerated the fixtures again in the 3rd commit.

For non-test conversions I think we should add a check on the Polymer version. Soon the new features we need will be release, and for workspace conversion we can specify #master. Ultimately we want to read conversion metadata from npm.

PR #255, opened 2017-12-19
use npm install instead of yarn install

FredKSchott:
See #254:

> There are a few issues uncovered with running `yarn install --flat` in a workspace:
> 
> 1. Yarn has a bug where the git hash of a locally referenced dependency (that's also been converted) is read as the package "version", causing errors inside of yarn ("83c19296e1cc615c224df685c497a8e5e38dcdf5 is not a valid version") .
> 2. The new, desired post-conversion version is written to each package.json, which can cause conflicts when installing (ex: `can't flatten semver ranges "3.0.0-pre.1" & "3.0.0-pre.2").

Reverting to use `npm install` now, but we should fix this so that we have proper support for yarn.

### DIFF
--- a/src/test-workspace.ts
+++ b/src/test-workspace.ts
@@ -35,13 +35,6 @@ async function setupRepos(
     reposUnderTest: WorkspaceRepo[],
     localConversionMap: ConversionResultsMap,
     options: WorkspaceTestSettings) {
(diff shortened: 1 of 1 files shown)


stramel:
Maybe this isn't the place to discuss this, but I'm curious if you guys are planning on moving forward with `yarn` or `npm`'s asset module whenever it is finished?

PR #257, opened 2017-12-21
Add importPath static property to elements

### LINKED ISSUE #97: Use import.meta.url
URLs in CSS were file-relative. After processing with html2js they're main document relative.

There's also the resolveUrl API, which does similar things.

What should we do here?

Failing polymer test to confirm:

unit/custom-style.html
  - style paths in included dom-modules loaded in import

also most of unit/resolveurl.html

justinfagnani:
Fixes #97

### DIFF
--- a/src/document-converter.ts
+++ b/src/document-converter.ts
@@ -245,6 +245,9 @@ export class DocumentConverter {
       // We need to inline templates on a per-script basis, otherwise we run
       // into trouble matching up analyzer AST nodes with our own.
       this.inlineTemplates(scriptProgram, script);
+      if (this.conversionSettings.addImportPath) {
+        this.addImportPathsToElements(scriptProgram, script);
+      }
       const comments: string[] = getCommentsBetween(
           this.document.parsedDocument.ast, prevScriptNode, script.astNode);
       const statements =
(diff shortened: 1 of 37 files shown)


src/document-converter.ts
```diff
@@ -691,6 +694,65 @@ export class DocumentConverter {
...
+              jsc.identifier('import'), jsc.identifier('meta')),
+          jsc.identifier('url'));
+
+      const node = nodePath.node;
+      if (node.type === 'ClassDeclaration' || node.type === 'ClassExpression') {
+        // A Polymer 2.0 class-based element
+        node.body.body.splice(
+            0,
+            0,
+            jsc.methodDefinition(
```
FredKSchott: pull this into a variable

src/document-converter.ts
```diff
@@ -691,6 +694,65 @@ export class DocumentConverter {
...
+                    null, [], jsc.blockStatement([jsc.returnStatement(
+                                  importMetaUrl)])),
+                true));
+      } else if (node.type === 'CallExpression') {
+        // A Polymer hybrid/legacy factory function element
+        const arg = node.arguments[0];
+        if (arg && arg.type === 'ObjectExpression') {
+          arg.properties.unshift(jsc.property(
+              'init', jsc.identifier('importPath'), importMetaUrl));
+        }
```
FredKSchott: else, emit warning

justinfagnani: done

src/document-converter.ts
justinfagnani: done

PR #261, opened 2018-01-05
Delete HTML files that are simple wrappers on external scripts

### LINKED ISSUE #200: modulizer is overwriting .js files
Given this project layout:

```
- iron-test-helpers/
  - test-helpers.html (just a script tag loading "iron-test-helpers.js")
  - test-helpers.js
```

modulizer returns this:
```
- iron-test-helpers/
  D test-helpers.html
  M test-helpers.js (contains only the (now broken) imports from orig. html file)
```

justinfagnani:
Fixes #200 

This is a pretty conservative version of a fix. It only effects files that only have a script tag that points to what would be name of the converted output file. We can extend this to eliminate wrappers that point to other files, but those don't always result in a file being overwritten like this case. We should also probably add detection of overwrites in general.

### DIFF
--- a/src/document-converter.ts
+++ b/src/document-converter.ts
@@ -236,6 +236,16 @@ export class DocumentConverter {
   }
 
   convertToJsModule(): ConversionResult {
+    if (this._isWrapperHTMLDocument) {
+      return {
+        originalUrl: this.originalUrl,
+        convertedUrl: this.convertedUrl,
+        convertedFilePath: getJsModuleConvertedFilePath(this.originalUrl),
+        deleteOriginal: true,
+        output: undefined,
+      };
+    }
+
     const combinedToplevelStatements = [];
     let prevScriptNode: parse5.ASTNode|undefined = undefined;
     for (const script of this.document.getFeatures({kind: 'js-document'})) {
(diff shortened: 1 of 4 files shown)


stramel:
Aww man, you got me excited @justinfagnani! I thought this was covering #182 lol

justinfagnani:
That can come next based off this, but here we were actually breaking conversions so I wanted to fix it first.

src/document-converter.ts
```diff
@@ -266,9 +276,9 @@ export class DocumentConverter {
     const importedReferences = this.collectNamespacedReferences(program);
     // Add imports for every non-module <script> tag to just import the file
     // itself.
-    for (const scriptImports of this.document.getFeatures(
+    for (const scriptImport of this.document.getFeatures(
```
FredKSchott: 👍

src/js-module.ts
```diff
@@ -32,14 +32,22 @@ export interface ConversionResult {
...
    * originalUrl, the original file will always be overwritten with the
    * converted output.
    */
   readonly deleteOriginal?: boolean;
-  readonly output: HtmlFile|JsModule;
+
+  /**
+   * An object describing the converted output of this module, or `undefined`.
+   *
+   * If `undefined`, the file is deleted regardless of the value of
```
FredKSchott: mention that no new file is created either, that was the 1 tricky big I got while reading this.

justinfagnani: ok

PR #262, opened 2018-01-05
Add a Polymer 3.0 Elements status doc

FredKSchott:


docs/polymer-3-element-status.md
```diff
@@ -0,0 +1,128 @@
+# Polymer 3.0 Status Page
+
+**Last run: 2018-01-05**
+
+This table contains the status of every Polymer 3.0 element being run through automated testing.
```
FredKSchott: I didn't want to spend too many cycles getting this explanation right until I'd had a chance to look at the blog post. We want it to match the same communication we use there.

justinfagnani: yep!

PR #267, opened 2018-01-08
Support specifying a git branch when setting up test fixtures.

aomarks:
Also makes update-fixtures script set a non-zero exit code if any of the fixtures fail to setup.

### DIFF
--- a/src/tools/update-fixtures.ts
+++ b/src/tools/update-fixtures.ts
@@ -87,7 +87,7 @@ async function overridePolymer(sourceDir: string) {
 }
 
 (async () => {
-  let errored = false;
+  let exitCode = 0;
 
   await Promise.all([
     updateFixture({
@@ -111,10 +111,8 @@ async function overridePolymer(sourceDir: string) {
   ].map((p) => p.catch((e) => {
     // Exit with an error code if any fixture fails, but let them all finish.
     console.error(e);
-    errored = true;
+    exitCode = 1;
   })));
 
-  if (errored) {
-    process.exit(1);
-  }
+  process.exit(exitCode);
 })();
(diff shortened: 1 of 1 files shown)


src/tools/update-fixtures.ts
```diff
@@ -37,13 +38,14 @@ async function updateFixture(options: UpdateFixtureOptions) {
...
+  const branch = options.branch || 'master';
 
-  console.log(`Cloning ${options.repoUrl} to ${sourceDir}...`);
+  console.log(`Cloning ${options.repoUrl} #${branch} to ${sourceDir}...`);
   await mkdirp(fixturesDir);
   rimraf.sync(sourceDir);
 
   await exec(
-      `git clone ${options.repoUrl} ${sourceDir} --depth=1`,
+      `git clone ${options.repoUrl} ${sourceDir} --branch=${branch} --depth=1`,
```
rictic: This looks like this requires the branch, but the branch arg is optional?

FredKSchott: it should be optional (so that we can use the default branch if none provided)

src/tools/update-fixtures.ts
```diff
@@ -85,6 +87,8 @@ async function overridePolymer(sourceDir: string) {
 }
 
 (async () => {
+  let errored = false;
```
FredKSchott: errored is a little vague, how about `exitCode`, set as 0, and then set as 1 in the catch handler

src/tools/update-fixtures.ts
aomarks: Right above, it defaults to `master`. Just seemed simpler than modifying this command.

src/tools/update-fixtures.ts
aomarks: Done.

src/tools/update-fixtures.ts
FredKSchott: ah right! Okay I'm happy with this, it doesn't support a different default branch, but that's a minor minor minor point.

PR #263, opened 2018-01-05
Avoid file name collisions from converted HTML files

justinfagnani:
Additional fix for #200

PR #269, opened 2018-01-10
Avoid file name collisions from converted HTML files

FredKSchott:
Supersedes #263, Improves #200

Improves on Justin's work in #263 to convert imported scripts that are loaded in a converted module.

### DIFF
--- a/src/document-converter.ts
+++ b/src/document-converter.ts
@@ -87,6 +87,19 @@ const legacyJavascriptTypes: ReadonlySet<string|null> = new Set([
 ]);
 
 /**
+ * This is a set of JavaScript files that we know to be converted as modules.
+ * This is neccessary because we don't definitively know the type of an external
+ * JavaScript file loaded in to a top-level HTML file.
+ *
+ * TODO(fks): Add the ability to know via conversion manifest support or
+ * convert dependencies as entire packages instead of file-by-file.
+ * (See https://github.com/Polymer/polymer-modulizer/issues/268)
+ */
+const knownScriptModules = new Set<string>([
+  'iron-test-helpers/mock-interactions.js',
+]);
+
+/**
  * Detect legacy JavaScript "type" attributes.
  */
 function isLegacyJavaScriptTag(scriptNode: parse5.ASTNode) {
@@ -235,6 +248,16 @@ export class DocumentConverter {
                 !this.conversionSettings.excludes.has(f.document.url));
   }
 
+  private isInternalNonModuleImport(scriptImport: Import): boolean {
+    const oldScriptUrl = getDocumentUrl(scriptImport.document);
+    const newScriptUrl = this.convertScriptUrl(oldScriptUrl);
+    const isModuleImport =
+        dom5.getAttribute(scriptImport.astNode, 'type') === 'module';
+    const isInternalImport =
+        this.urlHandler.isImportInternal(this.convertedUrl, newScriptUrl);
+    return isInternalImport && !isModuleImport;
+  }
+
   convertToJsModule(): ConversionResult {
     if (this._isWrapperHTMLDocument) {
       return {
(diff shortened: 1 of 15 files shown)


src/document-converter.ts
```diff
@@ -86,6 +86,19 @@ const legacyJavascriptTypes: ReadonlySet<string|null> = new Set([
   'text/x-javascript',
 ]);
 
+/**
+ * This is a set of JavaScript files that we know to be converted as modules.
+ * This is neccessary because we don't definitively know the type of an external
+ * JavaScript file loaded in to a top-level HTML file.
```
justinfagnani: We know the type on in the input - that's completely determined by the `module` attribute. I think you're saying we don't know they type of the output, yeah?

src/document-converter.ts
```diff
@@ -248,15 +271,29 @@ export class DocumentConverter {
 
     const combinedToplevelStatements = [];
     let prevScriptNode: parse5.ASTNode|undefined = undefined;
-    for (const script of this.document.getFeatures({kind: 'js-document'})) {
+    const allFeatures = this.document.getFeatures();
+    const relevantScripts = Array.from(allFeatures).filter((f) => {
```
justinfagnani: `relevantScripts` -> `scriptToConvert`|`scriptToInline`?

src/document-converter.ts
```diff
@@ -433,11 +474,19 @@ export class DocumentConverter {
       const offsets = htmlDocument.sourceRangeToOffsets(
           htmlDocument.sourceRangeForNode(scriptImport.astNode)!);
 
-      const correctedUrl = this.formatImportUrl(
-          this.convertDocumentUrl(getDocumentUrl(scriptImport.document)),
-          scriptImport.url);
+      const convertedUrl =
+          this.convertDocumentUrl(getDocumentUrl(scriptImport.document));
+      const correctedUrl = this.formatImportUrl(convertedUrl, scriptImport.url);
```
justinfagnani: corrected how?

src/document-converter.ts
```diff
@@ -248,15 +271,29 @@ export class DocumentConverter {
...
+        return true;
+      }
+      if (f.kinds.has('html-script')) {
+        return this.isInternalNonModuleImport(f as Import);
+      }
+      return false;
+    });
+
+    for (const script of relevantScripts) {
+      const scriptDocument = script.kinds.has('js-document') ?
```
justinfagnani: For the external scripts that don't have the same path as this document, we need to return a result that marks them for deletion.

FredKSchott: just clarifying: you mean external to the current file, and not to the current package, correct?

src/document-converter.ts
FredKSchott: good point, I'll clarify

src/document-converter.ts
justinfagnani: yes :)

basically, we should delete the files we inline and don't overwrite (or delete all inlined files as long as the order of writing results works out)

FredKSchott: got it, just making sure I understood you

FredKSchott:
@justinfagnani thanks for the review! PTAL (cleaned up the code a bit more as well)

FredKSchott:
Superseded by #270

src/document-converter.ts
```diff
@@ -274,13 +309,24 @@ export class DocumentConverter {
...
     // Add imports for every non-module <script> tag to just import the file
     // itself.
-    for (const scriptImport of this.document.getFeatures(
-             {kind: 'html-script'})) {
+    for (const scriptImport of this.document.getFeatures({ kind: 'html-script' })) {
       const oldScriptUrl = getDocumentUrl(scriptImport.document);
       const newScriptUrl = this.convertScriptUrl(oldScriptUrl);
-      importedReferences.set(newScriptUrl, new Set());
+      if (convertedHtmlScripts.has(scriptImport)) {
+        results.push({
```
justinfagnani: leave a comment that JS files with the same path as the convertedFilePath of _this_ file are deleted here, but re-written by the result pushed below, because the ordering of results matters.

PR #270, opened 2018-01-11
Add import references to a11ySuite in tests

FredKSchott:
Related to https://github.com/Polymer/web-component-tester/pull/665

`a11ySuite` is now a module, so we need to import references to it directly. This creates an explicit import `a11ySuite` vs. the more convoluted ESM->ESM exporting via the window object.


src/document-converter.ts
```diff
@@ -573,6 +576,10 @@ export class DocumentConverter {
     removeUnnecessaryEventListeners(program);
     removeWrappingIIFEs(program);
     const importedReferences = this.collectNamespacedReferences(program);
+    const wasA11ySuiteAdded = addA11ySuiteIfUsed(
```
justinfagnani: Can `a11Suite` be treated as a namespaced reference and handled in `collectNamespacedReferences`?

FredKSchott: As discussed, it could but it would require a bit more work since a11ySuite is a bit of a one-off. Adding a todo to address later (if manifest conversion doesn't fix it for us.)

PR #271, opened 2018-01-12
Update polymer-3-element-status.md

FredKSchott:


docs/polymer-3-element-status.md
```diff
@@ -14,115 +14,113 @@ This table contains the status of every Polymer 3.0 element being run through au
 
 ## Support Table
 
-| repo | `npm install` | `npm test` | reason? |
+| repo | *npm install* | *npm test* | reason? |
```
stramel: I wonder if it would be good to show the command used here instead or not tie it to the word npm?

FredKSchott: oh, looking at this now this was supposed to stay `npm install`, wonder if I changed it on purpose or by accident.

These are the commands run within each repo in question. Would `$ npm install` convey this more? Something else? Open to suggestions!

stramel: hmmm didn't notice that point until you mentioned it. I was thinking that you were using `yarn install` instead of `npm install` since AFAIK `flat: true` in the `package.json` won't do anything unless using `yarn` currently.

justinfagnani:
@FredKSchott should the be updated as of yesterday?

docs/polymer-3-element-status.md
FredKSchott: @stramel great point, I agree `| repo | `npm install` | `wct --npm` | reason? |`  makes the most sense here, with a note about why we're testing `npm install` vs. `yarn install` at the moment.

FredKSchott:
Merging this now so that we show the latest data as linked to from the blog post, can address table headers in future PR

PR #272, opened 2018-01-12
Add deleteFiles option and --delete-files flag.

### LINKED ISSUE #259: Delete or update TypeScript declaration files
The existing declarations are not written for modules.

aomarks:
It deletes all files/directories matching some glob patterns, after conversion.

Fixes #259 (assuming we remember to add `--delete-files types` when we convert Polymer).


src/convert-workspace.ts
```diff
@@ -106,6 +106,13 @@ export default async function convert(options: WorkspaceConversionSettings):
   // Process & write each conversion result:
   await writeFileResults(options.workspaceDir, converter.getResults());
 
+  // Delete files that were explicitly requested to be deleted.
+  for (const repo of options.reposToConvert) {
+    for (const glob of options.deleteFiles || []) {
+      await rimraf(path.join(repo.dir, glob));
```
justinfagnani: So this means that the globs apply to all repos, and we can't write a glob that just applies to one by prefixing the repo name? Fine either way, but clarify in a comment.

aomarks:
> Looks good, but to make sure we don't have a drive-deleting bug, can you update the integration tests to use the flag and update the fixtures? Since polymer@master has a types directory you can even try out the intended use case.

Not easily, because the integration tests only check the converted files output, so the types aren't there anyway (along with LICENSE, etc.). In general, should we be copying all non-converted files to the output directory?

src/convert-workspace.ts
aomarks: Done.

justinfagnani:
> Not easily, because the integration tests only check the converted files output, so the types aren't there anyway (along with LICENSE, etc.). In general, should we be copying all non-converted files to the output directory?

Argh, good point. I think maybe we want to copy the input dir, then run modulizer with `--out .`

aomarks:
> Argh, good point. I think maybe we want to copy the input dir, then run modulizer with --out .

Do you mean in general, or for the test framework?

justinfagnani:
> Do you mean in general, or for the test framework?

Right now I mean just for the integration tests.

Ultimately `--out` might not be worth it. Maybe we should just always destructively write to the package directory, assume git to rollback, and confirm the changes before writing to disk.

aomarks:
> Right now I mean just for the integration tests.

Done. Added an integration test that we delete the types directory. Depends on pending PR https://github.com/Polymer/polymer-modulizer/pull/278.

FredKSchott:
-1 for removing `--out`, at least for us as developers. Having to `git reset --hard` & `git clean -fd` after each run is a pain and git clean isn't exactly risk-free

PR #276, opened 2018-01-16
New: Update travis file

### LINKED ISSUE #186: Update .travis.yml files to use --npm and yarn
See https://github.com/PolymerElements/marked-element/pull/77

stramel:
A basic implementation to update travis file commands to work with P3.

Not sure if this is the route you guys want to take with this so I'm just putting it out here. Feel free to decline if it isn't what you guys want.

Note: Tested this with `marked-element` and seemed to work.

Fixes #186


src/project-converter.ts
```diff
@@ -203,4 +203,16 @@ export class ProjectConverter {
 
     return results;
   }
+
+  /**
+   * Updates polymer cli and wct usages to work with js modules
+   */
+  convertTravisYaml(travisYaml: string) {
+    return travisYaml
+        .replace(/polymer\ install.*/g, 'yarn')  // polymer install => yarn
```
justinfagnani: these replacements seem a little aggressive. Is there a way to limit them to the fields we expect to see them in, or at least to values, not keys?

stramel: @justinfagnani Yeah, I was thinking the same thing. When I started looking at more `.travis.yml` files, I had some open questions that maybe you can help out with?

stramel:
@justinfagnani Still need some input on the open questions above.

Tests look to be passing other than some yaml formatting in the expected. Will dig into soon.

justinfagnani:
> Will `polymer install --variants` convert to `polymer install --npm --variants`?

We haven't made a call on this, but I'm inclined to try to reduce our tool workload in the transition as much as possible. Installing variants of dependencies seems to be a package manager concern, and I think if it's critical to use on npm, we should lobby npm for that feature.

> I assume we will convert `bower install` to `polymer install --npm` or `yarn install`?

We haven't made this final call either. Even if we don't strictly require it, we'd strongly prefer flat installations. It may end up being `yarn install --flat`. I think this should be configurable with a flag.

I don't know of any conflicts the Travis install behavior will cause. I think we can wait and see on that.

stramel:
Consolidated into #351. Thanks @azakus!

PR #278, opened 2018-01-17
Make integration tests do in-place conversion by copying files first.

aomarks:
For the integration tests, we now copy the source dir to the output dir before running the conversion, so we are doing an in-place conversion. This will let us test https://github.com/Polymer/polymer-modulizer/pull/272 (the new delete option), and also confirms we aren't doing anything bad to non-converted files.

Also switches a couple files to `fs-extra`.

Definitely look at the first commit separately, the second commit just adds all the previously missing files to the expected output.

Depends on https://github.com/Polymer/polymer-modulizer/pull/277

Part of https://github.com/Polymer/polymer-modulizer/issues/259


src/test/fixtures/fixtures_test.ts
```diff
@@ -118,14 +117,14 @@ suite('Fixtures', () => {
...
-      test(`packages/${fixtureBasename}`, async () => {
+      test(`packages/${fixtureBasename}`, async() => {
         const fixtureSourceDir = path.join(fixtureDir, 'source');
         const fixtureExpectedDir = path.join(fixtureDir, 'expected');
         const fixtureResultDir = path.join(fixtureDir, 'generated');
         const fixtureTestConfig = require(path.join(fixtureDir, 'test.js'));
         assert.isOk(fs.statSync(fixtureSourceDir).isDirectory());
         assert.isOk(fs.statSync(fixtureExpectedDir).isDirectory());
-        rimraf.sync(fixtureResultDir);
+        fs.copy(fixtureSourceDir, fixtureResultDir);
```
FredKSchott: isn't this async? Add `await`

aomarks: Done.

> Can you confirm that this runs in any reasonable amount of time? I believe the current tests diff all files across expected and generated, so I expect this adds a ton of work to compare the two untouched bower_components directories.

After the bower change, `npm run test:full` runs in 35 sec, exactly the same as running it on `master`.

> Regardless, add logic to ignore diffing bower_components & any other known-okay dirs

Done. We now don't copy any `bower_components` or `node_modules` files.

src/test/fixtures/fixtures_test.ts
```diff
@@ -117,14 +115,18 @@ suite('Fixtures', () => {
...
-      test(`packages/${fixtureBasename}`, async() => {
+      test(`packages/${fixtureBasename}`, async () => {
         const fixtureSourceDir = path.join(fixtureDir, 'source');
         const fixtureExpectedDir = path.join(fixtureDir, 'expected');
         const fixtureResultDir = path.join(fixtureDir, 'generated');
         const fixtureTestConfig = require(path.join(fixtureDir, 'test.js'));
         assert.isOk(fs.statSync(fixtureSourceDir).isDirectory());
         assert.isOk(fs.statSync(fixtureExpectedDir).isDirectory());
-        fs.copy(fixtureSourceDir, fixtureResultDir);
+        await fs.emptyDir(fixtureResultDir);
```
FredKSchott: ah good catch!

src/tools/update-fixtures.ts
```diff
@@ -52,7 +52,11 @@ async function updateFixture(options: UpdateFixtureOptions) {
   await exec('bower install', {cwd: sourceDir});
 
   // We're going to do an in-place conversion.
-  await fs.copy(sourceDir, convertedDir);
+  await fs.emptyDir(convertedDir);
+  await fs.copy(sourceDir, convertedDir, {
```
FredKSchott: If we're doing an in-place conversion, we need the bower components in the `convertedDir` so that they're properly analyzed as a part of conversion. 

Are we sure we're doing an in-place conversion here? A few lines down I see:
```
await convertPackage({
   inDir: sourceDir,
  outDir: convertedDir,
```

Which looks like we're still reading one directory and writing to another.

aomarks: > If we're doing an in-place conversion, we need the bower components in the convertedDir so that they're properly analyzed as a part of conversion.

Done. Now I include bower_components, but I exclude them during diffing.

> Are we sure we're doing an in-place conversion here? A few lines down I see:
> 
> await convertPackage({
>    inDir: sourceDir,
>   outDir: convertedDir,
> Which looks like we're still reading one directory and writing to another.

Oops, yeah. I guess it worked anyway, but now it's a real in-place upgrade.

src/test/fixtures/fixtures_test.ts
```diff
@@ -123,27 +123,27 @@ suite('Fixtures', () => {
...
-        });
+        await fs.copy(fixtureSourceDir, fixtureResultDir);
 
         // Top-Level Integration Test! Test the CLI interface directly.
-        const output = await exec(fixtureSourceDir, 'node', [
+        const output = await exec(fixtureResultDir, 'node', [
           modulizerBinPath,
           '--out',
-          '../generated/',
+          fixtureResultDir,
```
FredKSchott: can you do `'.'` here instead? That would be more true to how our users use it, and less likely to get out of sync with the first argument.

aomarks: Done.

PR #279, opened 2018-01-17
escape JS comments inside js comments

FredKSchott:


src/document-util.ts
```diff
@@ -426,12 +426,13 @@ export function attachCommentsToFirstStatement(
   const firstStatement: (estree.Statement|estree.ModuleDeclaration)&RecastNode =
       statements[0]!;
 
-  const recastComments: RecastComment[] = comments.map((c) => {
+  const recastComments: RecastComment[] = comments.map((comment) => {
+    const escapedComment = comment.replace('*/', '*\\/');
```
FredKSchott: note that this only escapes the closing comment, leaving the opening comment untouched. This is syntactically all we need to fix the issue but for symmetry we may want to escape both. Open to thoughts here.

src/test/unit/project-converter_test.ts
```diff
@@ -2485,6 +2485,42 @@ console.log('second script');
...
+  /**
+   *  Final comment
+   **/
+-->`
+      });
+
+      assertSources(await convert(), {
+        'test.js': `
+// comment in script
+/* /* First comment *\\/ */
```
stramel: Seems odd that the first comment is now after the `// comment in script`

FredKSchott: Agreed. This is common across all our tests, I'll add a separate issue to tackle

FredKSchott: Agreed, filed as https://github.com/Polymer/polymer-modulizer/issues/281

src/document-util.ts
justinfagnani: +1

PR #284, opened 2018-01-17
run workspace tests repo-by-repo

FredKSchott:
When running modulizer on a giant batch of elements, currently we need to wait for the entire batch to complete before getting any error/success messages.

### DIFF
--- a/src/test-workspace.ts
+++ b/src/test-workspace.ts
@@ -47,28 +47,25 @@ async function setupRepos(
  * Run `yarn install` to install dependencies in a repo. Note that this creates
  * a node_modules/ folder & an associated yarn.lock file as side effects.
  */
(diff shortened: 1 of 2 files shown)


src/test-workspace.ts
```diff
@@ -105,26 +102,28 @@ export async function testWorkspace(
     localConversionMap: Map<string, string>, options: WorkspaceTestSettings) {
   const allRepos = options.reposToConvert;
 
-  logStep(1, 5, '🔧', `Preparing Repos...`);
+  logStep(1, 4, '🔧', `Preparing Repos...`);
```
justinfagnani: want to make that `4` a variable?

FredKSchott: Eh it's pretty minor, if we have to touch this again then we can do it

PR #290, opened 2018-01-20
Fix: version

### LINKED ISSUE #288: --version is broken
🙀

stramel:
@FredKSchott Fixes #288 

Also, ran format.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "test:watch": "watchy -w src/ -w custom_typings/ -- npm run test:fast --loglevel=silent",
     "test": "npm run test:fast && npm run lint",
     "lint": "tslint -p ./",
-    "format": "find src test | grep '\\.js$\\|\\.ts$' | xargs clang-format --style=file -i",
+    "format": "find bin src test | grep \"\\.js$\\|\\.ts$\" | xargs clang-format --style=file -i",
     "clean": "rm -rf ./lib/",
     "prepublishOnly": "npm test",
     "prepare": "npm run build"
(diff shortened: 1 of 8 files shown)


stramel:
> By default, `--max_old_space_size` (which controls the upper limit of the V8 heap) is ~1.5GB. If your system has less memory than that, you will need to explicitly set this option to the appropriate amount of memory available to that process.

Maybe it would be better to not have this on there by default and just add a note in the README?

FredKSchott:
That's a fair point, we set that as the default because we were running into it for the 100+ element runs we've been doing.

If your machine has less memory than 8192MB, what happens? I assumed it still ran and would only cause issues if you were only trying to use more memory than your machine had available, which would be the case in either case.

FredKSchott:
From @stramel in Slack:
> Yeah, so with the `--max-old-space-size=8192` it just spins out and uses my system's resources
> once removed i get the `../package.json` error

Okay then, I think that's safe to remove now. We're getting ready to bring our own process into a single executable file / shell script, so we can handle this there for our specific need instead

package.json
```diff
@@ -19,7 +19,7 @@
     "test:watch": "watchy -w src/ -w custom_typings/ -- npm run test:fast --loglevel=silent",
     "test": "npm run test:fast && npm run lint",
     "lint": "tslint -p ./",
-    "format": "find src test | grep '\\.js$\\|\\.ts$' | xargs clang-format --style=file -i",
+    "format": "find bin src test | grep \"\\.js$\\|\\.ts$\" | xargs clang-format --style=file -i",
```
stramel: This command will properly run on windows now too.

FredKSchott: nice!

FredKSchott:
Thanks @stramel!

PR #311, opened 2018-02-01
update analyzer to v3.0.0-pre.10, typescript

### LINKED ISSUE #306: Update analyzer from v2->v3

FredKSchott:
Fixes #306

Background: Polymer Analyzer v3 includes a major change to how urls are handled. Instead of being package-relative, they are now absolute to the file system. polymer-modulizer was built to work with package-relative urls, and gets a lot of benefits from that system over complete absolute 
urls (easier to check against, manipulate, etc). 

This PR updates the existing calls to `getDocumentUrl()` to handle the analyzer's new url format and get all the analyzer updates from the last few months. We now need to keep a reference to the analyzer during conversion itself, but it fits well into our existing concept of a conversion UrlHandler.

Long-term, we could consider doing the work to move to use the analyzer's absolute url format internally as well to take advantage of the analyzer's `relative` URL logic. But at the moment there's no pressing reason to doing so.

### DIFF
--- a/src/conversion-settings.ts
+++ b/src/conversion-settings.ts
@@ -15,7 +15,7 @@
 import * as estree from 'estree';
 import {Iterable as IterableX} from 'ix';
 import * as jsc from 'jscodeshift';
-import {Analysis} from 'polymer-analyzer';
+import {Analysis, Analyzer} from 'polymer-analyzer';
 
 export type NpmImportStyle = 'name'|'path';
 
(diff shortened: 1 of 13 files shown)


src/conversion-settings.ts
```diff
@@ -180,12 +180,15 @@ export function createDefaultConversionSettings(
...
+          .filter((imp) => !imp.kinds.has('html-script-back-reference'))
+          .map(
+              (imp) =>
+                  analyzer.urlResolver.relative(imp.url) as string | undefined)
           .filter(
-              (url) =>
-                  !(url.startsWith('bower_components') ||
-                    url.startsWith('node_modules')));
-  const includes = new Set(importedFiles);
+              (url) => !!url &&
```
justinfagnani: I prefer explicit checks so we know what values to expect here. `url !== undefined`?

PR #319, opened 2018-02-10
format all files with vscode

e111077:
this project includes a `.vscode/settings.json` which specifies format on save. This means that if I want to make a one-line change to a file, then it is muddled with all of these formatting changes.

Alternative to this PR is to simply set that flag to false.

### DIFF
--- a/custom_typings/ast-types.d.ts
+++ b/custom_typings/ast-types.d.ts
@@ -1,12 +1,12 @@
 
 declare module 'ast-types' {
   import * as estree from 'estree';
-  import {Node} from 'estree';
+  import { Node } from 'estree';
   export interface NodePath<N extends Node = Node> {
     node: N;
     parent?: NodePath;
 
-    get<NKey extends keyof N>(name: NKey): NodePath<N[NKey]>|undefined;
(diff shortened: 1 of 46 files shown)


stramel:
:+1: I ALWAYS have to run the formatter after I save files in vscode before I submit PR.

rictic:
We auto-format with clang-format. There's a clang-format vscode plugin, and we've got a script to run the formatter over all files that you can run with: `npm run format`

Maybe we should add clang-format to the vscode recommended plugins for this repo? https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions

stramel:
Probably would be good for any of the repos where you are using clang formatter to maintain the formatting and have the vscode settings.

TimvdLippe:
You can do that with `git diff --exitcode`. We use that in Polymer/Polymer: https://github.com/Polymer/polymer/pull/5003

stramel:
@timvdlippe I thinking you just add it to the precommit hook. So code doesn't ever get committed without the formatting being ran. Something like [Husky](https://github.com/typicode/husky) makes that really easy.

e111077:
got clang compiler. Closing PR opening issue for continuation of this discussion

PR #322, opened 2018-02-10
fix multiple js comments in a html comment

### LINKED ISSUE #321: CSS comments in HTML comments
CSS comments present in the HTML typically the file documentation seem to break modulizer.

see: https://github.com/PolymerElements/app-layout/blob/master/app-grid/app-grid-style.html#L84
result: https://github.com/PolymerElements/app-layout/blob/__auto_generated_3.0_preview/app-grid/app-grid-style.js#L163

e111077:
fixes #321

### DIFF
--- a/src/document-util.ts
+++ b/src/document-util.ts
@@ -428,7 +428,7 @@ export function attachCommentsToFirstStatement(
       statements[0]!;
 
   const recastComments: RecastComment[] = comments.map((comment) => {
-    const escapedComment = comment.replace('*/', '*\\/');
+    const escapedComment = comment.replace(/\*\//g, '*\\/');
     return {
       type: 'Block' as 'Block',
       leading: true,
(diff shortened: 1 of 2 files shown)


src/document-util.ts
```diff
@@ -428,7 +428,7 @@ export function attachCommentsToFirstStatement(
     statements[0]!;
 
   const recastComments: RecastComment[] = comments.map((comment) => {
-    const escapedComment = comment.replace('*/', '*\\/');
+    const escapedComment = comment.replace(new RegExp('\\*/', 'g'), '*\\/');
```
stramel: Not sure if this would be preferred to `new Regex`, `/\*\//g`?

FredKSchott:
From: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace  
> "Only the first occurrence will be replaced."

Doh! Good catch.

PR #327, opened 2018-02-12
Fix: Remove top-level use-strict statement

### LINKED ISSUE #181: Should we remove 'use-strict'?

stramel:
Fixes #181


FredKSchott:
@stramel awesome, good call. Lets split this out into a separate pass so that the logic doesn't exist inside the "top level this rewriting" pass.

stramel:
@FredKSchott I moved this to a separate pass can you review again, please?

stramel:
Looks good :+1:

PR #328, opened 2018-02-13
add missing deps marked and prism

e111077:
Fixes marked-element tests. ~Something else is also messing with prism-element~ fixed.

### DIFF
--- a/dependency-map.json
+++ b/dependency-map.json
@@ -240,8 +240,8 @@
     "semver": "^3.0.0-pre.8"
   },
   "prism": {
-    "npm": "prism",
-    "semver": "^4.0.0"
+    "npm": "prismjs",
+    "semver": "^1.11.0"
   },
   "prism-element": {
     "npm": "@polymer/prism-element",
--- a/dependency-map.json
+++ b/dependency-map.json
@@ -231,9 +231,17 @@
     "npm": "@polymer/iron-demo-helpers",
     "semver": "^3.0.0-pre.8"
   },
+  "marked": {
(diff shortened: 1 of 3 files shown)


dependency-map.json
```diff
@@ -231,10 +231,18 @@
...
+  "marked": {
+    "npm": "marked",
+    "semver": "~0.3.9"
+  },
   "marked-element": {
     "npm": "@polymer/marked-element",
     "semver": "^3.0.0-pre.8"
   },
+  "prism": {
+    "npm": "prism",
```
justinfagnani: shouldn't this be `prismjs`?

stramel:
I'm confused... isn't `marked` not setup to be a module or does that matter?

justinfagnani:
@stramel it might not matter. Some scripts actually work when loaded as modules. They still write to globals, and references to them won't be rewritten by modulizer. I've looked at Prism and I think it'll work. We'll see with marked.

dependency-map.json
e111077: Yes, yes it should

e111077:
@justinfagnani is right, the import just seems to load it as a script. Marked seems to work as well no problem

e111077:
That seemed to do the trick to fix prism-element. I'm impressed at how resilient the modulizer is already.

e111077:
investigating travis failure

justinfagnani:
@e111077 awesome!

You need to update the integration tests, which are expecting marked to not be found.

PR #329, opened 2018-02-14
Perform atomic package analysis/conversion before modifying files

FredKSchott:
## What is this?

This PR represents a big change to how we run conversions across files and packages. The main goals were to:
1. Perform repeatable package conversions that aren't dependent on how any one file is imported (current logic runs conversions file by file in a way that affects how a package is converted)
2. Package conversion should be represented by some serializable data that could be written to file and cached with each package. Future work will allow the modulizer to read this file instead of performing independent conversions of each dependency.
3. Unintentionally, making those changes involved splitting the current conversion process into two processes, which ~forced~ allowed us to also fix support for cyclical imports, which currently break the modulizer.

## Implementation tl;dr

- Packages are now converted as entire packages, instead of document-by-document. 
- This PR adds a new scanning stage into the conversion process. Before any package is converted to ESM, it's interface and the interface of all dependency packages are scanned.
- Scanning is a form of static analysis that doesn't block on cyclical imports.
- Splitting out this step of the conversion simplifies the conversion logic overall.

### DIFF
--- a/fixtures/packages/polymer/expected/test/smoke/html-tag.html
+++ b/fixtures/packages/polymer/expected/test/smoke/html-tag.html
@@ -16,7 +16,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <body>
   <script type="module">
 import { Element } from '../../polymer-element.js';
-import { html as html$0 } from '../../polymer.js';
+import { html as html$0 } from '../../lib/utils/html-tag.js';
 const html = html$0;
 class SuperClass extends Element {
   static get is() {return 'super-class';}
--- a/src/convert-workspace.ts
+++ b/src/convert-workspace.ts
@@ -58,7 +58,8 @@ function configureAnalyzer(options: WorkspaceConversionSettings) {
   const urlResolver = new PackageUrlResolver({packageDir: workspaceDir});
   const urlLoader = new InMemoryOverlayUrlLoader(new FSUrlLoader(workspaceDir));
   for (const [url, contents] of polymerFileOverrides) {
-    urlLoader.urlContentsMap.set(urlResolver.resolve(`polymer/${url}` as ResolvedUrl)!, contents);
+    urlLoader.urlContentsMap.set(
+        urlResolver.resolve(`polymer/${url}` as ResolvedUrl)!, contents);
   }
   return new Analyzer({
     urlLoader,
@@ -80,14 +81,14 @@ export default async function convert(options: WorkspaceConversionSettings):
     Promise<ConversionResultsMap> {
   const analyzer = configureAnalyzer(options);
   const analysis = await analyzer.analyzePackage();
-  const htmlDocuments = [...analysis.getFeatures({kind: 'html-document'})];
   const conversionSettings =
       createDefaultConversionSettings(analyzer, analysis, options);
   const urlHandler = new WorkspaceUrlHandler(analyzer, options.workspaceDir);
-  const converter = new ProjectConverter(urlHandler, conversionSettings);
-  const convertedPackageResults: ConversionResultsMap = new Map();
+  const converter =
+      new ProjectConverter(analysis, urlHandler, conversionSettings);
+  const scannedPackageResults: ConversionResultsMap = new Map();
 
-  // For each repo, convert the relevant HTML documents:
+  // For each repo, convert the full package:
   for (const repo of options.reposToConvert) {
     const repoDirName = path.basename(repo.dir);
     const bowerConfigPath = path.join(repo.dir, 'bower.json');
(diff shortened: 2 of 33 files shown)


src/project-converter.ts
```diff
@@ -43,41 +35,74 @@ const excludeFromResults = new Set([
...
+   */
+  private _convertPackage(
+      matchPackageName: string, visitedPackages: Set<string>) {
+    const htmlDocuments =
+        [...this.analysis.getFeatures({kind: 'html-document'})].filter((d) => {
+          const documentUrl = this.urlHandler.getDocumentUrl(d);
+          const packageName =
+              this.urlHandler.getOriginalPackageNameForUrl(documentUrl);
+          return packageName === matchPackageName;
+        });
```
justinfagnani: I would sort the files first for deterministic ordering: first by directory depth, then alpha.

justinfagnani: Actually, I would add a document-order traversal from that list to get the final ordering...

FredKSchott:
@justinfagnani @usergenic The PR is finally cleaned up and officially ready for review. PTAL

src/project-converter.ts
```diff
@@ -37,154 +38,98 @@ const excludeFromResults = new Set([
...
   /**
    * Convert a document and any of its dependencies. The output format (JS
    * Module or HTML Document) is determined by whether the file is included in
    * conversionSettings.includes.
    */
+  convertPackage(matchPackageName: string) {
+    // First, scan the package (and any of its dependencies)
+    this.scanner.scanPackage(matchPackageName);
+    // Then, convert each document in the given package
+    [...this.analysis.getFeatures({kind: 'html-document'})]
```
usergenic: Most of this `getFeatures` ... `filter` business is duplicated in `scanPackage`.  Maybe there's a way to DRY up the document query as `getHtmlDocumentsForPackage` and then call into it from `convertPackage` `scanPackage`?

FredKSchott: Great idea!

fixtures/packages/polymer/expected/test/smoke/html-tag.html
```diff
@@ -16,7 +16,7 @@
 <body>
   <script type="module">
 import { Element } from '../../polymer-element.js';
-import { html as html$0 } from '../../polymer.js';
+import { html as html$0 } from '../../lib/utils/html-tag.js';
```
justinfagnani: Add issue to get this to import from polymer.js

src/convert-workspace.ts
```diff
@@ -80,34 +81,28 @@ export default async function convert(options: WorkspaceConversionSettings):
...
-      const documentUrl = urlHandler.getDocumentUrl(document);
-      if (!documentUrl.startsWith(repoDirName) ||
-          conversionSettings.excludes.has(documentUrl)) {
-        continue;
-      }
-      converter.convertDocument(document);
-    }
-    convertedPackageResults.set(npmPackageName, repo.dir);
+    scannedPackageResults.set(npmPackageName, repo.dir);
+    converter.convertPackage(repoDirName);
```
justinfagnani: 👍

src/document-converter.ts
```diff
@@ -128,6 +128,24 @@ interface Edit {
   replacementText: string;
 }
 
+export interface JsModuleScanResult {
```
justinfagnani: comment these

src/document-converter.ts
```diff
@@ -292,7 +304,11 @@ export class DocumentConverter {
       rewriteToplevelThis(scriptProgram);
       // We need to inline templates on a per-script basis, otherwise we run
       // into trouble matching up analyzer AST nodes with our own.
-      this.inlineTemplates(scriptProgram, scriptDocument);
+      const localClaimedDomModules =
+          this.inlineTemplates(scriptProgram, scriptDocument);
+      for (const claimedDomModule of localClaimedDomModules) {
```
justinfagnani: 🏌️: `localClaimedDomModules.forEach` is nice, if you want

src/document-converter.ts
```diff
@@ -362,17 +410,26 @@ export class DocumentConverter {
...
-        es6Exports: new Set(exportMigrationRecords.map((r) => r.es6ExportName))
-      }
+      output: outputProgram.code + EOL
     });
     return results;
   }
 
-  convertAsToplevelHtmlDocument(): ConversionResult {
+  /**
+   * Scan a document as a top-level HTML document.
```
justinfagnani: add an explanation of why this doesn't do any work

src/project-scanner.ts
```diff
@@ -0,0 +1,179 @@
...
+    this.conversionSettings = conversionSettings;
+  }
+
+  /**
+   * Get all relevant HTML documents from a package that should be scanned,
+   * coverted, or otherwise handled by the modulizer.
+   */
+  getPackageDocuments(matchPackageName: string) {
+    return [...this.analysis.getFeatures({kind: 'html-document'})]
+        .filter((d) => d.isInline === false)
```
justinfagnani: Each filter does a pass through the array, so combine these into one.

fixtures/packages/polymer/expected/test/smoke/html-tag.html
FredKSchott: Tracking as #336

PR #330, opened 2018-02-15
updated passing tests

e111077:
4 now green; lots more passing tests


e111077:
Hmm this is actually missing some of my updates. All the good elements are passing

stramel:
@e111077 Yeah, I just checked the `gold-phone-input` and was passing. Good to see more stuff passing! :+1:

PR #332, opened 2018-02-15
Update passing elements

e111077:
Continuation of #330. (I forgot to hit ⌘+S)

10 more passing elements (14 if you add the 4 from #330) 
Lots more passing tests


stramel:
:tada:

PR #335, opened 2018-02-21
Merge existing package.json when generating the new one.

### LINKED ISSUE #334: Merge existing package.json for newly generated packages
Related: https://github.com/Polymer/polymer-modulizer/issues/251

aomarks:
Fixes #334

Also:

- Added `--flat` and `--private` CLI flags, both of which default to `false`. Previously we set `flat` to `true` by default, and did not touch `private`.
- Map Bower `main` foo.html to NPM `main` foo.js. Previously we just copied  it.
- Check for multiple licenses (valid for Bower, not for NPM) and warn.
- Add typings for NPM and Bower configs.
- Map Bower `authors` to NPM `author`/`contributors`. Previously we directly copied `authors`, which doesn't exist in Bower (although we use it in at least one element, so I kept it as a fallback).
- Improve warning messages.

### DIFF
--- a/custom_typings/bower-config.d.ts
+++ b/custom_typings/bower-config.d.ts
@@ -1,0 +1,23 @@
+// https://github.com/bower/spec/blob/master/json.md
+interface BowerConfig {
+  name: string;
+  description?: string;
+  version?: string;
+  main?: string|string[];
+  license?: string|string[];
+  ignore?: string[];
+  keywords?: string[];
+  authors?: Array<BowerPerson|string>;
+  homepage?: string;
+  repository?: {type: string, url: string};
+  dependencies?: {[pkg: string]: string};
+  devDependencies?: {[pkg: string]: string};
+  resolutions?: {[pkg: string]: string};
+  private?: boolean;
+}
+
+interface BowerPerson {
+  name?: string;
+  email?: string;
+  homepage?: string;
+}
--- a/src/convert-package.ts
+++ b/src/convert-package.ts
@@ -16,14 +16,15 @@ import * as fse from 'fs-extra';
 import * as path from 'path';
 import {Analysis, Analyzer, FSUrlLoader, InMemoryOverlayUrlLoader, PackageUrlResolver, ResolvedUrl} from 'polymer-analyzer';
 
+import {BowerConfig} from './bower-config';
 import {ConversionSettings, createDefaultConversionSettings, PartialConversionSettings} from './conversion-settings';
 import {generatePackageJson, writeJson} from './manifest-converter';
+import {YarnConfig} from './npm-config';
 import {ProjectConverter} from './project-converter';
 import {polymerFileOverrides} from './special-casing';
 import {PackageUrlHandler} from './urls/package-url-handler';
 import {PackageType} from './urls/types';
-import {mkdirp, rimraf, writeFileResults} from './util';
-
+import {mkdirp, readJsonIfExists, rimraf, writeFileResults} from './util';
 
 /**
  * Configuration options required for package-layout conversions. Contains
@@ -116,8 +117,6 @@ function configureAnalyzer(options: PackageConversionSettings) {
  */
 export default async function convert(options: PackageConversionSettings) {
   const outDir = options.outDir;
-  const npmPackageName = options.packageName;
-  const npmPackageVersion = options.packageVersion;
   await setupOutDir(outDir, options.cleanOutDir);
 
   // Configure the analyzer and run an analysis of the package.
@@ -156,17 +155,19 @@ export default async function convert(options: PackageConversionSettings) {
   }
 
   const packageJsonPath = path.join(options.inDir, 'package.json');
-  let existingPackageJson: Partial<YarnConfig>|undefined;
-  if (await fse.pathExists(packageJsonPath)) {
-    existingPackageJson = await fse.readJSON(packageJsonPath);
-  }
+  const existingPackageJson =
+      await readJsonIfExists<Partial<YarnConfig>>(packageJsonPath);
 
   // Generate a new package.json, and write it to disk.
   try {
     const packageJson = generatePackageJson(
         bowerJson,
-        npmPackageName,
-        npmPackageVersion,
+        {
+          name: options.packageName,
+          version: options.packageVersion,
+          flat: options.flat,
+          private: options.private,
+        },
         undefined,
         existingPackageJson);
     writeJson(packageJson, packageJsonPath);
--- a/src/convert-package.ts
+++ b/src/convert-package.ts
@@ -12,11 +12,12 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import * as fse from 'fs-extra';
 import * as path from 'path';
 import {Analysis, Analyzer, FSUrlLoader, InMemoryOverlayUrlLoader, PackageUrlResolver, ResolvedUrl} from 'polymer-analyzer';
 
 import {ConversionSettings, createDefaultConversionSettings, PartialConversionSettings} from './conversion-settings';
-import {generatePackageJson, readJson, writeJson} from './manifest-converter';
+import {generatePackageJson, writeJson} from './manifest-converter';
 import {ProjectConverter} from './project-converter';
 import {polymerFileOverrides} from './special-casing';
 import {PackageUrlHandler} from './urls/package-url-handler';
@@ -102,8 +103,7 @@ function configureAnalyzer(options: PackageConversionSettings) {
   for (const [url, contents] of polymerFileOverrides) {
     urlLoader.urlContentsMap.set(urlResolver.resolve(url)!, contents);
     urlLoader.urlContentsMap.set(
-        urlResolver.resolve(`../polymer/${url}` as ResolvedUrl)!,
-        contents);
+        urlResolver.resolve(`../polymer/${url}` as ResolvedUrl)!, contents);
   }
   return new Analyzer({
     urlLoader,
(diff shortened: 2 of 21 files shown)


src/convert-package.ts
```diff
@@ -12,11 +12,12 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import * as fse from 'fs-extra';
 import * as path from 'path';
 import {Analysis, Analyzer, FSUrlLoader, InMemoryOverlayUrlLoader, PackageUrlResolver, ResolvedUrl} from 'polymer-analyzer';
 
 import {ConversionSettings, createDefaultConversionSettings, PartialConversionSettings} from './conversion-settings';
-import {generatePackageJson, readJson, writeJson} from './manifest-converter';
+import {generatePackageJson, writeJson} from './manifest-converter';
```
stramel: Is there a benefit to maintaining the `writeJson` instead of using `fse.writeJson`?

aomarks: It converts `\n` to `os.EOL`. `fse.writeJson` actually supports this, but it's not in the latest typings. I might send a PR to add it to the typings. @FredKSchott Do we actually need to do this `\n` to `EOL` conversion for windows?

stramel: @aomarks Without the `\n` to `EOL` conversions, it will cause the expected fixtures test to fail due to size differences.

aomarks: Gotcha, makes sense. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23831

custom_typings/bower-config.d.ts
```diff
@@ -0,0 +1,23 @@
+// https://github.com/bower/spec/blob/master/json.md
+interface BowerConfig {
```
justinfagnani: If these are just global, and not typing a particular module, I'm not sure they should be in custom_typings. They can just be a .ts file in src/

src/manifest-converter.ts
```diff
@@ -92,55 +93,155 @@ export function writeJson(json: any, ...pathPieces: string[]) {
...
+      if (bowerJson.license.length === 1) {
+        license = bowerJson.license[0];
+      } else {
+        console.warn(
+            `${bowerJson.name}: Found multiple licenses in bower.json, ` +
+            `but package.json must have only one.`);
+      }
+    }
+    if (license) {
+      if (license.includes('polymer.github.io/LICENSE')) {
```
FredKSchott: Follow https://docs.npmjs.com/files/package.json#license / https://yarnpkg.com/lang/en/docs/package-json/#toc-license and use `license:  "SEE LICENSE AT polymer.github.io/LICENSE"`

src/convert-workspace.ts
```diff
@@ -42,12 +43,23 @@ export const GIT_STAGING_BRANCH_NAME = 'polymer-modulizer-staging';
...
+  const bowerJson = fse.readJSONSync(bowerJsonPath) as Partial<BowerConfig>;
   const npmPackageName =
       lookupNpmPackageName(bowerJsonPath) || bowerPackageName;
-  const packageJson =
-      generatePackageJson(bowerJson, npmPackageName, packageVersion);
-  writeJson(packageJson, repo.dir, 'package.json');
+
+  const packageJsonPath = path.join(repo.dir, 'package.json');
+  let existingPackageJson: Partial<YarnConfig>|undefined;
+  if (fse.pathExistsSync(packageJsonPath)) {
```
justinfagnani: This loading code is basically duplicated. Consider putting it in the generate function.

src/manifest-converter.ts
```diff
@@ -92,55 +93,155 @@ export function writeJson(json: any, ...pathPieces: string[]) {
...
-    resolutions: {
-      'inherits': '2.0.3',
-      'samsam': '1.1.3',
-      'supports-color': '3.1.2',
-      'type-detect': '1.0.0',
-    }
+
+    ...existingPackageJson,
+
+    name,
```
justinfagnani: maybe we should warn if the name doesn't match?

src/manifest-converter.ts
```diff
@@ -92,55 +93,155 @@ export function writeJson(json: any, ...pathPieces: string[]) {
...
+    version,
+    flat: true,
```
justinfagnani: Bonus: put this behind a flag. Will be useful with other upcoming features.

src/manifest-converter.ts
```diff
@@ -92,55 +93,155 @@ export function writeJson(json: any, ...pathPieces: string[]) {
...
+        console.warn(
+            `${bowerJson.name}: Found multiple mains in bower.json, ` +
+            `but package.json must have only one.`);
+      }
+    }
+    if (main && main.endsWith('.html')) {
+      // Assume that the bower main is already a correct relative path to an
+      // HTML file, and that the module equivalent will be in the same directory
+      // but with a JS extension.
+      packageJson.main = replaceHtmlExtensionIfFound(main);
```
justinfagnani: Add a TODO to use the conversion result to look up the new path.

src/manifest-converter.ts
```diff
@@ -92,55 +93,155 @@ export function writeJson(json: any, ...pathPieces: string[]) {
...
+      console.warn(
+          `${bowerJson.name}: ` +
+          `Could not automatically find appropriate license. ` +
+          `Please manually set your package.json license according to ` +
+          `https://docs.npmjs.com/files/package.json#license`);
+    }
   }
 
-  for (const bowerPackageName in bowerJson.dependencies) {
+  packageJson.dependencies = Object.assign({}, packageJson.dependencies);
```
justinfagnani:
```ts
packageJson.dependencies = {...packageJson.dependencies)};
```

src/manifest-converter.ts
```diff
@@ -92,55 +93,155 @@ export function writeJson(json: any, ...pathPieces: string[]) {
...
+  for (const bowerPackageName in bowerJson.dependencies || []) {
     setNpmDependencyFromBower(
         packageJson.dependencies, bowerPackageName, useLocal);
   }
-  for (const bowerPackageName in bowerJson.devDependencies) {
+  packageJson.devDependencies = Object.assign({}, packageJson.devDependencies);
```
justinfagnani: here too

src/manifest-converter.ts
```diff
@@ -92,55 +93,155 @@ export function writeJson(json: any, ...pathPieces: string[]) {
...
+        //
+        // There could also be multiple HTML files in main, e.g. a repo like
+        // paper-behaviors which contains 3 separate Polymer behaviors. We
+        // currently let that be, so importing the module directly will fail.
+        // We could also generate an index that re-exports all of the symbols
+        // from all of the mains.
+        console.warn(
+            `${bowerJson.name}: Found multiple mains in bower.json, ` +
+            `but package.json must have only one.`);
+      }
```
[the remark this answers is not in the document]
aomarks: Is it okay? It would mean that when somebody requires your module by name without adding `/actual-module`, an error would occur.

src/manifest-converter.ts
aomarks: Our license is identical to the standard 3-Clause BSD at https://opensource.org/licenses/BSD-3-Clause except that we use the word "owner" instead of "holder" in one place, and we format the list with stars instead of numbers. Pretty annoying, I wonder why we made these trivial differences. Should we really link to our version, or just use the standard SPDX ID? I feel like using the standard ID is going to help with automation.

aomarks: (Also in case it wasn't clear, this was the previous behavior, I did not change this behavior in this PR.)

stramel: I previously had implemented the conversion. I didn't realize you could do `"SEE LICENSE IN" at that time.

src/manifest-converter.ts
```diff
@@ -92,55 +93,155 @@ export function writeJson(json: any, ...pathPieces: string[]) {
...
+
+    ...existingPackageJson,
+
+    name,
+    version,
+    flat: true,
+    // TODO(aomarks) We probably want this behind a flag, because switching an
+    // existing package to be publishable could be dangerous. We need this for
+    // the Polymer elements, because the 2.x branches are marked private.
+    private: undefined,
```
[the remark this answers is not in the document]
aomarks: Added a flag.

src/manifest-converter.ts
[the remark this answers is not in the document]
aomarks: Reworked this section PTAL.

src/manifest-converter.ts
```diff
@@ -92,55 +93,155 @@ export function writeJson(json: any, ...pathPieces: string[]) {
...
+            `but package.json must have only one.`);
+      }
+    }
+    if (main && main.endsWith('.html')) {
+      // Assume that the bower main is already a correct relative path to an
+      // HTML file, and that the module equivalent will be in the same directory
+      // but with a JS extension.
+      packageJson.main = replaceHtmlExtensionIfFound(main);
+    } else {
+      console.warn(
```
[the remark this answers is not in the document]
aomarks: Done.

custom_typings/bower-config.d.ts
aomarks: Turned them into modules in `src/`.

src/convert-workspace.ts
aomarks: I added `util.readJsonIfExists<T>` LMKWYT! Didn't really feel we should to add an FS dependency to this function.

src/manifest-converter.ts
aomarks: Done.

src/manifest-converter.ts
aomarks: Done.

src/manifest-converter.ts
aomarks: Done.

src/manifest-converter.ts
aomarks: Done.

src/manifest-converter.ts
aomarks: Done.

aomarks:
> Can you finish the move readJson -> fse.readJSONSync and remove our custom readJson impl?

Done.

src/manifest-converter.ts
aomarks: We discussed offline and decided to keep the BSD-3-Clause ID.

PR #337, opened 2018-02-22
update to pre.10

e111077:


stramel:
I'm just curious, is there a reason to keep bumping these instead of using the `next` tag?

FredKSchott:
@stramel That might not be a bad move while we're iterating fast like this. In general writing release tags like `next` directly into a package.json is dangerous because a tag is constantly changing. But for these early preview temporary releases that doesn't really matter.

@e111077 LGTM, fixing conflicts then merging

PR #338, opened 2018-02-23
Update fixtures_test.ts to use `compareContent: true`

FredKSchott:
Blocked by #337


aomarks:
Rebased. This is passing on CI now because the latest commit updated the expecteds.

PR #339, opened 2018-02-23
Decouple updating test source repos from regenerating expected results.

aomarks:
Previously, the `build:fixtures` NPM script would both pull the latest master commit for each repo, and regenerate the expected result.

Now, you can run `npm run build:fixtures -- --skip-source-update` to re-generate the expected without also pulling in the latest source SHA This makes it easier to separate the effect of modulizer changes vs changes to the source repo, and is also faster to run during development.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
   },
   "scripts": {
     "build": "tsc",
-    "build:fixtures": "npm run build && node lib/tools/update-fixtures",
+    "update-fixture-sources": "npm run build && node lib/tools/update-fixture-sources",
+    "update-fixture-expected": "npm run build && UPDATE_EXPECTED=true mocha lib/test/fixtures/fixtures_test.js",
     "test:full": "npm run build && bash test/run-mocha.sh && npm run lint",
     "test:fast": "npm run build && bash test/run-mocha.sh unit",
     "test:watch": "watchy -w src/ -w custom_typings/ -- npm run test:fast --loglevel=silent",
--- a/package.json
+++ b/package.json
@@ -13,8 +13,7 @@
   },
   "scripts": {
     "build": "tsc",
-    "update-fixture-sources": "npm run build && node lib/tools/update-fixture-sources",
-    "update-fixture-expected": "npm run build && UPDATE_EXPECTED=true mocha lib/test/fixtures/fixtures_test.js",
+    "build:fixtures": "npm run build && node lib/tools/update-fixtures",
     "test:full": "npm run build && bash test/run-mocha.sh && npm run lint",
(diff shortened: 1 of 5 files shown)


src/tools/update-fixture-sources.ts
```diff
@@ -97,21 +74,14 @@ async function overridePolymer(sourceDir: string) {
     updateFixture({
       folder: 'polymer',
       repoUrl: 'https://github.com/Polymer/polymer.git',
-      packageName: '@polymer/polymer',
```
justinfagnani: is this removed because we have package.json's in the repos now?

aomarks: No, because this script no longer rubs modulizer at all so it doesn't need it.

package.json
```diff
@@ -13,7 +13,8 @@
   },
   "scripts": {
     "build": "tsc",
-    "build:fixtures": "npm run build && node lib/tools/update-fixtures",
+    "update-fixture-sources": "npm run build && node lib/tools/update-fixture-sources",
```
FredKSchott: Match existing form of scripts that do `verb:description`, something like `fetch:fixtures`/`generate:fixtures`/`build:fixtures`

src/test/fixtures/fixtures_test.ts
```diff
@@ -135,6 +135,12 @@ suite('Fixtures', () => {
           '--add-import-path',
         ].concat(fixtureTestConfig.options || []));
 
+        if (process.env.UPDATE_EXPECTED === 'true') {
```
FredKSchott: It feels weird to run a tool through an existing test, that actually generates its own fixture, and then become a noop test. If the purpose is to DRY up our fixture generation logic, could we instead move this logic into a shared place that both generation tool & tests could pull from?

package.json
aomarks: How is `update:fixture-sources` and `update:fixture-expected`? I find `build` a kind of unclear verb for this case, but if you prefer it I don't mind. To me the important distinction is sources vs expected output.

aomarks: Nm, now I have a flag to the existing update-fixtures script, so you can just do `npm run build:fixtures --skip-source-update`.

src/test/fixtures/fixtures_test.ts
aomarks: Ok, I added a new library and depend on it in both cases, PTAL.

I've used the "record mode" pattern I had before -- I kind of like it because it's so simple, it makes it harder to accidentally get your golden generation out of sync from the test itself, and you can use mocha's `--grep` flag to easily control which test golden you are re-generating. But I concede that it is hacky.

PR #340, opened 2018-02-24
--delete-files more safely, and don't check in expected/bower_components

aomarks:
Also:

- Update howto doc and test scripts to delete `**/*.d.ts` for the elements.
- Don't bother checking in `<foo>/expected/bower_components`. Our test logic always ignores bower_components, and it makes reading the diffs of fixture updates much harder.
- Bump the stdout buffer size for our `git status -s` command, because it blows up with large diffs.


stramel:
@aomarks WRT `git status -s` I think maybe switching to something like `git diff-index --quiet HEAD --` would be a better solution. It will just exit one if the index is dirty instead of printing something out.

src/util.ts
```diff
@@ -108,3 +110,26 @@ export async function readJsonIfExists<T>(filepath: string):
   }
   return undefined;
 }
+
+/**
+ * Delete all files matching any of the given glob patterns, rooted in the given
+ * directory, excluding any file in node_modules/ or bower_components/.
+ */
+export async function deleteGlobsSafe(
+    globs: Iterable<string>, cwd: string): Promise<void> {
```
justinfagnani: I try to be paranoid with accepting `Iterable` and explicitly rule out strings, especially with `Iterable<string>` since a string is one and even the type system won't help you out if there's a mixup.

aomarks: Modulizer uses `Iterable<string>` a lot in its configuration interfaces, so to do this consistently would mean changing a bunch of other places too: https://github.com/Polymer/polymer-modulizer/blob/master/src/conversion-settings.ts#L106

src/cli/command-package.ts
```diff
@@ -26,7 +26,9 @@ export default async function run(options: CliOptions) {
 
   // Ok, we're updating a package in a directory not under our control.
   // We need to be sure it's safe.
-  const {stdout, stderr} = await exec(inDir, 'git', ['status', '-s']);
+  const {stdout, stderr} = await exec(inDir, 'git', ['status', '-s'], {
+    maxBuffer: 1024 * 1024,  // The file list can get pretty long.
```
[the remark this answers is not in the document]
aomarks: It's a good idea but 1) this command doesn't cover untracked files, so we'd need to run two commands (I tried a bit to find a single command that would cover everything `git status` does, but doesn't find one), and 2) it's a bit awkward with `execFile` to check for non-zero exit code, because it throws, and you need to exclude more serious errors like file-not-found. Could totally do this, just doesn't seem worth the work.

Also, I noticed that we were running `git status` even when `force` was true, and then ignoring the result. So I re-arranged the `if` statement to not run the command at all with `force`, which was `true` in the code path I was hitting (updating the test fixtures).

aomarks:
Also added `bower.json` to the set of files we delete (offline LGTM).

PR #343, opened 2018-02-27
Scan exports by package, resolve export conflicts by project

FredKSchott:
Still trying to unblock package manifest reading/writing, but unfortunately lots of little known issues are blocking us. This PR gets us 99% there, up to the writing of the created manifest as a per-package scan artifact. Blocking the manifest writing is:

- **package mode**: Now that we're scanning packages instead of specific documents, the current project-level `includes` file set no longer works. We have a workaround for workspace mode, but we either need to bring that same workaround to package mode, remove package mode, or create a permanent fix now based on bower.json.
- **all modes:** files URLs are project-relative. We need to add package-relative URL logic so that URLs written in manifests are all package-relative. We currently don't have an idea of "package-relative" creation, or conversion from a bower- or npm-world URL.

I'm creating this PR as a checkpoint, then planning to fix these two blocking issues, then finally connecting the manifest to the "writeJSON" logic.

### DIFF
--- a/src/package-scanner.ts
+++ b/src/package-scanner.ts
@@ -1,0 +1,192 @@
+/**
+ * @license
+ * Copyright (c) 2018 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {Analysis, Document} from 'polymer-analyzer';
+
+import {ConversionSettings} from './conversion-settings';
+import {DeleteFileScanResult, DocumentConverter, HtmlDocumentScanResult, JsModuleScanResult} from './document-converter';
+import {JsExport} from './js-module';
+import {OriginalDocumentUrl} from './urls/types';
+import {UrlHandler} from './urls/url-handler';
+
+export interface PackageScanResult {
+  files:
+      Map<OriginalDocumentUrl,
+          JsModuleScanResult|DeleteFileScanResult|HtmlDocumentScanResult>;
+  exports: Map<string, JsExport>;
+}
+
+/**
+ * PackageScanner provides the top-level interface for scanning any single
+ * package. Scanning packages allows us to detect the new ES Module
+ * external interface(s) across a project so that we can properly rewrite and
+ * convert our files.
+ */
+export class PackageScanner {
+  private readonly packageName: string;
+  private readonly analysis: Analysis;
+  private readonly urlHandler: UrlHandler;
+  private readonly conversionSettings: ConversionSettings;
+
+  /**
+   * A set of all external dependencies (by name) actually detected as JS
+   * imported by this package.
+   */
+  readonly externalDependencies = new Set<string>();
+
+  /**
+   * All JS Exports for a single package registered by namespaced identifier,
+   * to map implicit HTML imports to explicit named JS imports.
+   */
+  private readonly namespacedExports = new Map<string, JsExport>();
+
+  /**
+   * All Scan Results for a single package registered by document URL, so that
+   * the conversion process knows how to treat each file.
+   */
+  private readonly results = new Map<
+      OriginalDocumentUrl,
+      JsModuleScanResult|DeleteFileScanResult|HtmlDocumentScanResult>();
+
+  constructor(
+      packageName: string, analysis: Analysis, urlHandler: UrlHandler,
+      conversionSettings: ConversionSettings) {
+    this.packageName = packageName;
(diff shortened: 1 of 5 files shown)


stramel:
I was thinking it would be good to remove package mode and modify workspaces to be given an existing directory? This would also bring the bower install, post nom install, and test checks to existing directories.

FredKSchott:
Yea agreed, I think the whole team is slowly coming together on that idea. Maintaining two different modes is just too expensive, but we do still want to support the quick-iterating-on-a-local-package workflow if we can.

src/package-scanner.ts
```diff
@@ -0,0 +1,192 @@
...
+        }
+      }
+    }
+  }
+
+  /**
+   * Check if a document is explicitly excluded or has already been scanned
+   * to decide if it should be skipped.
+   */
+  private _shouldScanDocument(document: Document): boolean {
```
justinfagnani: Let's be consistent on private names. I disagree with the Google TS style guide, since `private` is stripped away in compilation, but 🤷‍♂️

src/package-scanner.ts
```diff
@@ -0,0 +1,192 @@
...
+    const documentUrl = this.urlHandler.getDocumentUrl(document);
+    const packageName =
+        this.urlHandler.getOriginalPackageNameForUrl(documentUrl);
+
+    for (const htmlImport of DocumentConverter.getAllHtmlImports(document)) {
+      const importDocumentUrl = this.urlHandler.getDocumentUrl(<any>htmlImport);
+      const importPackageName =
+          this.urlHandler.getOriginalPackageNameForUrl(importDocumentUrl);
+
+      if (importPackageName === packageName) {
```
justinfagnani: It'd be nice to put the external check into a reusable function

src/project-scanner.ts
```diff
@@ -12,14 +12,31 @@
...
+  for (const [k, v] of exportsMap) {
+    exportsObject[k] = { name: v.name, url: v.url };
+  }
+  return exportsObject;
+}
+
+function filesMapToObject(filesMap: Map<OriginalDocumentUrl, ScanResult>) {
+  const filesObject: any = {};
+  for (const [k, v] of filesMap) {
+    filesObject[k] = (<any>v).convertedUrl || null;
```
justinfagnani: why the `any` cast? If it's really needed, use `as` syntax

src/package-scanner.ts
```diff
@@ -0,0 +1,192 @@
...
+import {DeleteFileScanResult, DocumentConverter, HtmlDocumentScanResult, JsModuleScanResult} from './document-converter';
+import {JsExport} from './js-module';
+import {OriginalDocumentUrl} from './urls/types';
+import {UrlHandler} from './urls/url-handler';
+
+export interface PackageScanResult {
+  files:
+      Map<OriginalDocumentUrl,
+          JsModuleScanResult|DeleteFileScanResult|HtmlDocumentScanResult>;
+  exports: Map<string, JsExport>;
```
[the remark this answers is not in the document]
FredKSchott: This is removed in the most recent PR

src/package-scanner.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * Get all relevant HTML documents from a package that should be scanned,
+   * coverted, or otherwise handled by the modulizer.
+   */
+  getPackageDocuments() {
+    return [
+      ...this.analysis.getFeatures(
+          {kind: 'html-document', externalPackages: true})
```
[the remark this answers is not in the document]
FredKSchott: We call this function on all packages, including dependencies of the main package-under-conversion.

src/package-scanner.ts
```diff
@@ -0,0 +1,192 @@
...
+      const packageName =
+          this.urlHandler.getOriginalPackageNameForUrl(documentUrl);
+      return packageName === this.packageName;
+    });
+  }
+
+  /**
+   * Return the results of the package scan.
+   */
+  getResults(): PackageScanResult {
```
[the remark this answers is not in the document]
FredKSchott: called by ProjectScanner

src/package-scanner.ts
```diff
@@ -0,0 +1,192 @@
...
+      exports: this.namespacedExports,
+    };
+  }
+
+  /**
+   * Scan a document and any of its dependency packages. The scan document
+   * format (JS Module or HTML Document) is determined by whether the file is
+   * included in conversionSettings.includes.
+   */
+  private scanDocument(document: Document, forceJs = false) {
```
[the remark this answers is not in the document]
FredKSchott: We don't capitalize full acronyms in lowerCameCase, happy to discuss that but lets to change in this PR

src/package-scanner.ts
FredKSchott: yea, I have mixed feelings here as well. Sticking without the leading `_` for now so that our codebase is consistent.

src/project-scanner.ts
FredKSchott: This is also removed in the next PR

FredKSchott:
The next PR fixes the issues introduced here. Instead of merging this immediately into `master` and introducing those issues to the codebase, the plan is to:

1. close this PR.
2. review the related PR.
3. merge that PR into this branch.
4. move the full, completed branch into master.

PR #346, opened 2018-03-01
handle package-relative URLs, add manifests in workspace mode

FredKSchott:
Write conversion manifests in workspace mode!

### DIFF
--- a/src/convert-workspace.ts
+++ b/src/convert-workspace.ts
@@ -63,6 +63,19 @@ async function writePackageJson(
   writeJson(packageJson, packageJsonPath);
 }
 
+
+/**
+ * For a given repo, generate a new package.json and write it to disk.
+ */
+async function writeConversionManifest(
+    repo: WorkspaceRepo, converter: ProjectConverter) {
+  const bowerPackageName = path.basename(repo.dir);
+  const manifestJsonPath = path.join(repo.dir, 'manifest.json');
+  const packageManifest = converter.getPackageManifest(bowerPackageName);
+  writeJson(packageManifest, manifestJsonPath);
+}
+
+
 /**
  * Configure a basic analyzer instance for the workspace.
  */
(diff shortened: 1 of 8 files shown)


src/urls/url-handler.ts
```diff
@@ -33,4 +33,9 @@ export interface UrlHandler {
       string;
   convertUrl(url: OriginalDocumentUrl): ConvertedDocumentUrl;
   createConvertedUrl(partialUrl: string): ConvertedDocumentUrl;
+
+  originalUrlToPackageRelative(url: OriginalDocumentUrl): string;
+  convertedUrlToPackageRelative(url: ConvertedDocumentUrl): string;
+  convertedDocumentFilePathToPackageRelative(url: ConvertedDocumentFilePath):
```
FredKSchott: The current system of project-specific URL handling is definitely hitting a breaking point here, we keep needing to add one-off workarounds to each specific URL handler.

I think we have agreement about *wanting* to move to a single workspace-mode conversion. I'm going to experiment with how easy it would be to preserve single-package conversion inside that layout, by doing something like cloning the local git repo at it's current branch into the workspace.

src/convert-workspace.ts
```diff
@@ -63,6 +63,19 @@ async function writePackageJson(
...
 }
 
+
+/**
+ * For a given repo, generate a new package.json and write it to disk.
+ */
+async function writeConversionManifest(
+    repo: WorkspaceRepo, converter: ProjectConverter) {
+  const bowerPackageName = path.basename(repo.dir);
+  const manifestJsonPath = path.join(repo.dir, 'manifest.json');
```
bicknellr: 'manifest.json' seems a bit generic; maybe this should include 'modulizer' in the name ('modulizer.manifest.json')?

FredKSchott:
Closing this in favor of a PR that addresses both reading and writing at the same time. (Will link here when created)

PR #347, opened 2018-03-02
Read a settings object from a comment, use to rename the file before conversion.

bicknellr:
This PR causes modulizer to read a specially formatted comment containing a JSON string that contains conversion settings specific to that file. The comment must start with `polymer-modulizer:` and the rest of the comment must be parseable as JSON. The options accepted are in the `FileConversionSettings` interface, `usedFileName` being the only one at the moment. `usedFileName` causes modulizer to act as though the file had been renamed to the given value just before converting. #4805 specifically mentions wanting the current 'polymer.html' to become 'polymer-legacy.js' after conversion, you could do this by inserting this comment into that file:

```html
<!--polymer-modulizer:{
  "usedFileName": "polymer-legacy.html"
}-->
```

The reason I decided to make this JSON object was that #28, which I'm planning to work on after this one, looks like it will need something similar to indicate how exports should be renamed:

```html
<!--polymer-modulizer:{
  "renameExports": {
    "Element": "PolymerElement"
  }
}-->
```

- [x] Bug: AFAICT, all HTML documents are having their extensions rewritten to `.js`, even if they aren't converted to a module.
- [x] Bug: All analyzed files, including dependencies, are being rewritten.
- [x] Tests


src/document-converter.ts
```diff
@@ -269,6 +300,38 @@ export class DocumentConverter {
...
+   * Finds the comment containing the FileConversionSettings object and returns
+   * the parsed object. Throws if more than one comment could be parsed as the
+   * FileConversionSettings object.
+   */
+  private getFileConversionSettings(document: ParsedHtmlDocument):
+      Partial<FileConversionSettings> {
+    const treeAdapter = parse5.treeAdapters.default;
+
+    let settings: Partial<FileConversionSettings>|null = null;
+    document.visit([(node: parse5.ASTNode) => {
```
bicknellr: I think visiting every node here is what's slowing everything down. I'm going to look around in the analyzer to see if it already found these comments.

bicknellr:
Talked w/ @FredKSchott about #346 and it seems like that PR should go in before this one. The performance problem here seems to be around traversing all of the files to find the comment so we can determine what each file was renamed to, but #346 captures all of this information in the manifest that it writes so these packages wouldn't need to be scanned again if that information is already available. Also, it seems like it would be better to go with a per-package input file that describes all of the file-specific changes in it (file / export renaming) so that files in the package being converted wouldn't need to be traversed either.

bicknellr:
I've merged #354 into this branch because I intend to add to and use the results of the conversion manifest. As of right now, I haven't made any local changes to account for the changes from that branch, so this is back in a WIP state. I'll comment again when this is closer.

justinfagnani:
@bicknellr since this code should just be determining how a name is interpreted, does it really rely on #354?

justinfagnani:
`usedFileName` reads a bit weird to me. How about `renameTo`?

bicknellr:
The name the user should provide is what this PR pretends the file's name was before conversion but `renameTo` sounds like it would be the name after conversion. I didn't want the provided name to be the after-conversion name so that it could have its extension potentially changed using the same logic as everything else. Specifically, I wanted to prevent the user from having to guess how the conversion process would normally change the file's name. I agree that `usedFileName` is awkward though. `renameFrom`?

(I'll leave another comment once we've had the offline discussion about the relation w/ #354.)

bicknellr:
#354 was merged to master and I merged master back into this branch, so the changes shown are now the intended content of the PR. This still needs some work though.

justinfagnani:
@bicknellr seems like there's still conflicts. Rebase?

bicknellr:
> conflicts

I don't see any at the moment.

FredKSchott:
(me trying to use the github dashboard for reviews) add me back when this is ready for re-review!

FredKSchott:
@justinfagnani refresh your page? I also see no conflicts

screen shot 2018-03-13 at 11 17 25 am

justinfagnani:
Odd. I've refreshed several times.

src/document-converter.ts
```diff
@@ -1073,18 +1143,55 @@ export class DocumentConverter {
...
+   * Converts this HTML Document's path from old world to new.
+   */
+  private generateSelfDocumentUrl(): ConvertedDocumentUrl {
+    let selfUrl = this.originalUrl;
+
+    // Replace the file name with `usedFileName` from this file's
+    // FileConversionSettings object, if provided.
+    const newName = this.fileConversionSettings.usedFileName;
+    if (typeof newName === 'string') {
+      const parts = selfUrl.split('/');
```
justinfagnani: why the split? Can a file not be moved between directories? Add comments

src/document-converter.ts
```diff
@@ -252,16 +272,23 @@ export class DocumentConverter {
...
+  private readonly fileConversionSettings: Partial<FileConversionSettings>;
+  private readonly document: Document<ParsedHtmlDocument>;
 
   constructor(
-      document: Document, urlHandler: UrlHandler,
-      conversionSettings: ConversionSettings) {
+      document: Document<ParsedHtmlDocument>, urlHandler: UrlHandler,
+      conversionSettings: ConversionSettings, selfScanResult?: ScanResult) {
     this.conversionSettings = conversionSettings;
+    this.fileConversionSettings =
```
justinfagnani: Shouldn't this be read in scanDocument?

src/document-converter.ts
```diff
@@ -273,6 +300,41 @@ export class DocumentConverter {
     return [...document.getFeatures({kind: 'html-import'})];
   }
 
+  private *
+      dom5DepthFirst(node: parse5.ASTNode): IterableIterator<parse5.ASTNode> {
+    const treeAdapter = parse5.treeAdapters.default;
+    yield node;
```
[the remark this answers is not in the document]
FredKSchott: also, we have a document-util file for document traversal helper functions like this

bicknellr:
Closing. @FredKSchott and I are working on a different approach to this, the first of which was #364.

bicknellr:
(Current work is happening in https://github.com/Polymer/polymer-modulizer/tree/rename-polymer.)

PR #349, opened 2018-03-03
Pass extra args given to `npm run` down to mocha for all test scripts.

bicknellr:
Extra arguments passed to `npm run` after the `--` are passed down to mocha. This lets you use features of mocha like filtering (`npm run test:full -- -g 'packages/iron-icon'` to test only the iron-icon fixture) or waiting for a debugger to attach (`npm run test -- --inspect-brk`).

The `fn() { ...; }; fn` part is because npm passes the args to the script by appending them to the end of the script; this way `fn` is called with those arguments and `$@` can be used to reference them inside `fn`.

For example, if you have a script like this:
```
"some-script": "target-for-args && clean-up"
```
... then `npm run-script some-script -- args here` runs:
```
target-for-args && clean-up args here
```
If the script is this, instead:
```
"some-script": "fn() { target-for-args \"$@\" && clean-up; }; fn"
```
... then `npm run-script some-script -- args here` effectively runs:
```
target-for-args args here && clean-up
```


bicknellr:
This seems to break in appveyor's shell.

bicknellr:
Oh... AppVeyor is for Windows... I'll have to rethink this.

PR #354, opened 2018-03-08
read & write conversion manifests from npm as a part of conversion

FredKSchott:
- Finalize the format for the conversion manifest
- Serialize a `PackageScanResult` to a format that can be written out with the package (the conversion manifest)
- De-serialize a `PackageScanResultJson` JSON manifest to a `PackageScanResult` internal representation, that is used to inform conversions.


src/package-scanner.ts
```diff
@@ -69,26 +74,84 @@ export class PackageScanner {
...
+    for (const document of this.getPackageHtmlDocuments()) {
       this.scanDocument(document);
     }
+  }
 
-    return this.getResults();
+  /**
+   * Fetch a conversion manifest from NPM. If none can be found, return null.
+   */
+  async getResultsFromManifest(): Promise<PackageScanResult|null> {
```
justinfagnani: null -> undefined

src/package-scanner.ts
```diff
@@ -69,26 +74,84 @@ export class PackageScanner {
...
    * coverted, or otherwise handled by the modulizer.
    */
-  getPackageDocuments() {
+  getPackageHtmlDocuments() {
     return [
-      ...this.analysis.getFeatures(
-          {kind: 'html-document', externalPackages: true})
+      ...this.analysis.getFeatures({
+        // Set externalPackages=true so that this method works on dependencies
+        // as well.
```
justinfagnani: this comment and the one below don't exactly say why we include external here and exclude it in the filter

src/package-scanner.ts
```diff
@@ -69,26 +74,84 @@ export class PackageScanner {
...
+   * Fetch a conversion manifest from NPM. If none can be found, return null.
+   */
+  async getResultsFromManifest(): Promise<PackageScanResult|null> {
+    const npmPackageInfo = lookupDependencyMapping(this.packageName);
+    if (!npmPackageInfo) {
+      return null;
+    }
+
+    try {
+      const unpkgResponse = await fetch(`https://unpkg.com/${
```
justinfagnani: parameterize this fetch call so this can be tested with a mock manifest fetcher

src/test/unit/conversion-manifest_test.ts
```diff
@@ -0,0 +1,76 @@
...
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {assert} from 'chai';
+import {readJson} from 'fs-extra';
+import * as path from 'path';
```
usergenic: I'm starting to get in the habit of importing `*` less.  `import {join} from 'path';`

Just a thought!

src/test/unit/project-converter_test.ts
```diff
@@ -72,7 +73,9 @@ suite('AnalysisConverter', () => {
     async function convert(
         partialOptions: Partial<TestConversionOptions> = {}) {
       // Extract options & settings /w defaults.
-      const packageName = partialOptions.packageName || 'some-package';
+      const bowerPackageName =
+          partialOptions.bowerPackageName || 'some-package';
+      const npmPackageName = partialOptions.npmPackageName || 'some-package';
```
usergenic: Another option:
```ts
const {
  npmPackageName = 'some-package',
  bowerPackageName = 'some-package',
  packageType = 'element',
  expectedWarnings = [],
} = partialOptions;
```

src/urls/package-url-handler.ts
```diff
@@ -45,11 +48,17 @@ export class PackageUrlHandler implements UrlHandler {
   }
 
   constructor(
-      analyzer: Analyzer, packageName: string,
-      packageType: PackageType = 'element') {
+      analyzer: Analyzer, bowerPackageName: string, npmPackageName: string,
+      packageType: PackageType = 'element', packageDir: string) {
     this.analyzer = analyzer;
-    this.packageName = packageName;
+    this.bowerPackageName = bowerPackageName;
```
usergenic: Sine the constructor is just assigning readonly props you can drop the declarations from the head of class and do them *inline* in constructor like this...
```ts
constructor(
  public readonly analyzer: Analyzer,
  public readonly bowerPackageName: string,
  public readonly npmPackageName: string,
  public readonly packageType: PackageType = 'element',
  public readonly packageDir: string) { }
```

src/urls/workspace-url-handler.ts
```diff
@@ -139,4 +144,38 @@ export class WorkspaceUrlHandler implements UrlHandler {
   getNameImportUrl(url: ConvertedDocumentUrl): ConvertedDocumentUrl {
     return url.slice('./'.length) as ConvertedDocumentUrl;
   }
+
+  originalUrlToPackageRelative(url: OriginalDocumentUrl): string {
+    return url.split('/').splice(1).join('/');
+  }
+
+  convertedUrlToPackageRelative(url: ConvertedDocumentUrl): string {
```
usergenic: Are these unit tested anywhere?

src/urls/package-url-handler.ts
bicknellr: I've always found this syntax confusing.

src/conversion-manifest.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Convert a single file scan result to a serializable JSON object.
+ */
+function serializePackageFileScanResult(
+    fileScanResult: ScanResult,
+    exportsMap: PackageScanExports,
+    urlHandler: UrlHandler): PackageFileJson|null {
+  if (fileScanResult.type === 'delete-file') {
+    return null;
```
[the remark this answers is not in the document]
FredKSchott: We actually want null here since undefined isn't serialized into JSON

src/conversion-manifest.ts
```diff
@@ -0,0 +1,154 @@
...
+  return fileExports;
+}
+
+/**
+ * Convert a single file scan result to a serializable JSON object.
+ */
+function serializePackageFileScanResult(
+    fileScanResult: ScanResult,
+    exportsMap: PackageScanExports,
+    urlHandler: UrlHandler): PackageFileJson|null {
```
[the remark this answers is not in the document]
FredKSchott: Same here

src/conversion-manifest.ts
```diff
@@ -0,0 +1,154 @@
...
+  if (fileScanResult.type === 'delete-file') {
+    return null;
+  }
+  const {convertedFilePath} = fileScanResult;
+  const convertedRelativePath =
+      urlHandler.convertedDocumentFilePathToPackageRelative(convertedFilePath);
+  if (fileScanResult.type === 'html-document') {
```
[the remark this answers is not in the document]
FredKSchott: I prefer the explicit three returns-by-type here for this function, vs. one return above and then a conditional return here. Let me know if you feel strongly about this.

src/conversion-manifest.ts
```diff
@@ -0,0 +1,154 @@
...
+  return {files: filesObject};
+}
+
+function fileMappingToScanResult(
+    originalUrl: OriginalDocumentUrl,
+    convertedUrl: ConvertedDocumentUrl|null,
+    fileData: PackageFileJson): ScanResult {
+  if (!convertedUrl) {
+    return {
+      type: 'delete-file',
```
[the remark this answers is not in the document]
FredKSchott: Again, I think there's enough complexity here to be very explicit about what type returns what format. Conditional return statements are much harder for me to read, personally.

src/conversion-manifest.ts
```diff
@@ -0,0 +1,154 @@
...
+    convertedPackageName: string,
+    conversionManifest: PackageScanResultJson,
+    urlHandler: UrlHandler): [PackageScanFiles, PackageScanExports] {
+  const filesMap: PackageScanFiles = new Map();
+  const exportsMap: PackageScanExports = new Map();
+  for (const [relativeFromUrl, fileData] of Object.entries(
+           conversionManifest.files)) {
+    const originalUrl = urlHandler.packageRelativeToOriginalUrl(
+        originalPackageName, relativeFromUrl);
+    const convertedUrl = fileData === null ?
```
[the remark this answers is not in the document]
FredKSchott: same here, comes from JSON so it needs null

src/package-scanner.ts
FredKSchott: Can we save this for when we add a mock manifest fetcher? Happy to add a TODO here if so. Or are you saying we should add tests that mock this out in this PR?

src/conversion-manifest.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+/**
+ * Convert a map of files to a serializable JSON object.
+ */
+export function serializePackageScanResult(
+    filesMap: PackageScanFiles,
+    exportsMap: PackageScanExports,
+    urlHandler: UrlHandler): PackageScanResultJson {
+  const filesObject: PackageFilesJson = {};
```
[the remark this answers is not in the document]
FredKSchott: i like it

src/conversion-manifest.ts
```diff
@@ -0,0 +1,154 @@
...
+        originalPackageName, relativeFromUrl);
+    const convertedUrl = fileData === null ?
+        null :
+        urlHandler.packageRelativeToConvertedUrl(
+            convertedPackageName, fileData.url);
+    filesMap.set(
+        originalUrl,
+        fileMappingToScanResult(originalUrl, convertedUrl, fileData!));
+  }
+  for (const [_originalUrl, scanResult] of filesMap) {
```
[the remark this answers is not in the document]
FredKSchott: Wahhhhhh? TIL

src/test/unit/conversion-manifest_test.ts
FredKSchott: I know, this is a hard habit for me to break. But I like the signal when I see `path.join` that I can assume it's the `path.join()` method that I already know, without investigating.

src/urls/package-url-handler.ts
FredKSchott: Agreed, I prefer as written ¯\_(ツ)_/¯

src/urls/workspace-url-handler.ts
FredKSchott: No :( I'll add some basic tests 👍

src/conversion-manifest.ts
```diff
@@ -0,0 +1,154 @@
...
+      convertedFilePath: getHtmlDocumentConvertedFilePath(originalUrl),
+    };
+  }
+  return {
+    type: 'js-module',
+    originalUrl: originalUrl,
+    convertedUrl: convertedUrl,
+    convertedFilePath: getJsModuleConvertedFilePath(originalUrl),
+    exportMigrationRecords:
+        Object.entries(fileData.exports).map(([exportId, exportName]) => ({
```
justinfagnani: for happier formatting, consider storing this in a variable first.

src/conversion-manifest.ts
```diff
@@ -0,0 +1,154 @@
...
+        originalPackageName, relativeFromUrl);
+    const convertedUrl = fileData === null ?
+        undefined :
+        urlHandler.packageRelativeToConvertedUrl(
+            convertedPackageName, fileData.convertedUrl);
+    filesMap.set(
+        originalUrl,
+        fileMappingToScanResult(originalUrl, convertedUrl, fileData!));
+  }
+  for (const [, scanResult] of filesMap) {
```
justinfagnani: nit: `for (const scanResult of filesMap.values()) {`

PR #357, opened 2018-03-13
add ability to map custom bower->npm dependencies

FredKSchott:
It's now possible to add custom dependency mappings for conversion.

Example:

```
modulizer --repo XXX YYY ZZZ --dependency-mapping old-name,new-name,^2.0.0 prism,prismjs,^1.0.0
```

### DIFF
--- a/src/cli/command-package.ts
+++ b/src/cli/command-package.ts
@@ -18,7 +18,10 @@ import * as semver from 'semver';
 
 import {CliOptions} from '../cli';
 import convertPackage from '../convert-package';
+import {saveDependencyMapping} from '../package-manifest';
 import {exec, logStep} from '../util';
+
+import {parseDependencyMappingInput} from './util';
 
 export default async function run(options: CliOptions) {
   const inDir = path.resolve(options.in || process.cwd());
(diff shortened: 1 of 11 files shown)


FredKSchott:
@justinfagnani okay ready for a look

src/cli/index.ts
```diff
@@ -96,6 +96,15 @@ const optionDefinitions: commandLineArgs.OptionDefinition[] = [
         'patterns.',
     defaultValue: []
   },
+  {
+    name: 'dependency-mapping',
+    type: String,
+    multiple: true,
+    description: 'A set of mapping instructions to map unknown bower ' +
+        'dependencies to npm. Must be of the format "bower,npm,semver".' +
```
justinfagnani: note that the version is the npm version

src/cli/command-package.ts
```diff
@@ -34,6 +37,17 @@ export default async function run(options: CliOptions) {
     process.exit(1);
   }
 
+  for (const rawMapping of options['dependency-mapping']) {
```
justinfagnani: can you move this whole thing into the parse function?

FredKSchott: I had this in an earlier version, but I liked the separation of concerns as-is. One method parses, one method saves, and then this for loop iterates it all over our CLI-provided options array. 

Happy to move this all into an additional `processAllDependencyMappings` function if you feel strongly here. Since you approved this I'll assume it's okay to merge as-is, but just ping me here / on chat and I'll add this in a separate PR.

PR #365, opened 2018-03-15
Add node_modules to gitignore

### LINKED ISSUE #183: Add node_modules to gitignore file
We should probably update the `.gitignore` file to have `node_modules`

azakus:
Fixes #183

### DIFF
--- a/src/gitignore.ts
+++ b/src/gitignore.ts
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright (c) 2018 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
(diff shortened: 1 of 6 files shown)


src/gitignore.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as fse from 'fs-extra';
+import {EOL} from 'os';
+
+const nodeModulesLine = 'node_modules';
+const searchNodeModulesLine = new RegExp(`^${nodeModulesLine}`);
```
justinfagnani: I've seen a number of variations used, like `/node_modules` to limit to top level. Maybe do `^$\/?{nodeModulesLine}`?

azakus: sure, will do

PR #367, opened 2018-03-16
rename 'polymer.html' to 'polymer-legacy.js', export `Polymer.Element` as `PolymerElement`

### LINKED ISSUE #28: Proposal: `Polymer.Element` should not be exported as `Element`
Please compare the following import statements:

1. `import { Element as PolymerElement } from 'polymer-element.js'`
vs
2. `import { PolymerElement } from 'polymer-element.js'`;

Some questions I might ask as a new user:
Why should I rename `Element` to `PolymerElement` (surely the Polymer authors would have named it that if they wanted it to be that)?
How is `Element` different from `window.Element`?
Why is a class called `Element` appearing twice in my element's ancestry?

It seems like the thing I'm expressing as a user is more concise (no `as` required) with #2, and even if some of the questions apply, I'm less inclined to ask them because `PolymerElement` is _obviously_ different from `Element`.

bicknellr:
Fixes #28, Polymer/polymer#4805.


src/document-converter.ts
```diff
@@ -1098,6 +1113,15 @@ export class DocumentConverter {
           'shadycss/custom-style-interface.html',
           'shadycss/entrypoints/custom-style-interface.js');
     }
+    if (this.originalPackageName === 'polymer') {
+      if (jsUrl === './polymer.html') {
+        jsUrl = jsUrl.replace('polymer.html', 'polymer-legacy.html');
```
justinfagnani: You don't really need a replace here, just `jsUrl = './polymer-legacy.html';`

justinfagnani: Why `polymer-legacy.html` and not `polymer-legacy.js`?

src/document-converter.ts
```diff
@@ -1098,6 +1113,15 @@ export class DocumentConverter {
...
+      }
+      if (jsUrl.endsWith('polymer/polymer.html')) {
+        jsUrl = jsUrl.replace(
```
justinfagnani: The replace is a bit loose compared to the endsWith check. Add an EOL anchor: `jsUrl = jsUrl.replace(/\.html$/, '-legacy.html')`

src/test/unit/project-converter_test.ts
```diff
@@ -452,6 +452,80 @@ export const Polymer = function(info) {
...
+          </script>
+        `,
+          });
+          const options = {
+            bowerPackageName: 'polymer',
+            npmPackageName: '@polymer/polymer',
+          };
+          assertSources(await convert(options), {
+            'test.js': `
+import { Polymer, foo } from \'./polymer-legacy.js\';
```
justinfagnani: You shouldn't need to escape quotes (`\'`) in a template literal.

src/test/unit/project-converter_test.ts
```diff
@@ -452,6 +452,80 @@ export const Polymer = function(info) {
       });
     });
 
+    test(
+        'in a package "polymer", renames a root file "/polymer.html" to "polymer-legacy.js"',
+        async () => {
+          setSources({
+            'test.html': `
+          <link rel="import" href="./polymer.html">
```
justinfagnani: does this also work for `<link rel="import" href="polymer.html">`, or non-sibling files?

bicknellr:
I'm going to address the comments here in next but we also shouldn't merge this until a PR based off of https://github.com/Polymer/polymer-modulizer/compare/wip-package-relative-includes is in so that we don't break package mode while fixing this in workspace mode.

src/document-converter.ts
```diff
@@ -248,20 +248,31 @@ function getImportDeclarations(
...
   private readonly document: Document;
 
   constructor(
-      document: Document, urlHandler: UrlHandler,
+      document: Document, originalPackageName: string, urlHandler: UrlHandler,
       conversionSettings: ConversionSettings) {
+    // The `originalPackageName` given by `PackageConverter` is always the name
+    // of the root package being converted, even if this `DocumentConverter` is
+    // converting a file in a dependency of the root package.
+    this.originalPackageName = originalPackageName;
```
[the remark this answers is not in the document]
bicknellr: Sorry, I realize now that this comment sounds like it's pointing out the opposite of what was intended: it's actually meant to describe a bug further up (in `PackageScanner`?), not state the intended value of `originalPackageName`. (i.e. What @FredKSchott is working on in the `wip-package-relative-includes` branch.)

src/document-converter.ts
bicknellr: The `replaceHtmlExtensionIfFound(jsUrl)` after these changes does the extension swap.

justinfagnani:
Ok, let me know when there's a PR for that.

src/test/unit/project-converter_test.ts
bicknellr: Fixed.

src/document-converter.ts
```diff
@@ -405,7 +416,9 @@ export class DocumentConverter {
         results.push({
           originalUrl: oldScriptUrl,
           convertedUrl: newScriptUrl,
-          convertedFilePath: getJsModuleConvertedFilePath(oldScriptUrl),
+          convertedFilePath:
+              replaceHtmlExtensionIfFound(oldScriptUrl as string) as
```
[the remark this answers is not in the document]
bicknellr: (from offline) We're going to make a similar function to `getJsModuleConvertedFilePath` that performs a similar cast but is specifically documented as intended to produce the final file name with renaming applied; as opposed to `getJsModuleConvertedFilePath`, which produces a `ConvertedFilePath` without considering renaming. This works for this specific case because the only file we are currently renaming is 'polymer.html': `convertScriptUrl` is only applied to `OriginalDocumentUrl`s that were already referencing files that were already separate scripts so it won't result in a renamed file. If we intend to make it possible to rename any file (e.g. with a special comment) then we'll need to change the new function to read from the renaming information produced during scanning.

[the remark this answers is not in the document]
bicknellr: Updated.

bicknellr:
Lots of converted dependencies are being written into the output directory - still investigating.

src/urls/package-url-handler.ts
```diff
@@ -240,7 +247,11 @@ export class PackageUrlHandler implements UrlHandler {
   }
 
   packageRelativeConvertedUrlToConvertedDocumentFilePath(
-      _originalPackageName: string, url: string): ConvertedDocumentFilePath {
-    return url as ConvertedDocumentFilePath;
+      originalPackageName: string, url: string): ConvertedDocumentFilePath {
+    if (originalPackageName === this.bowerPackageName) {
+      return url as ConvertedDocumentFilePath;
+    } else {
+      return `./bower_components/${originalPackageName}/${url}` as ConvertedDocumentFilePath;
```
bicknellr: This should be a reference into `node_modules` not `bower_components`.

bicknellr:
PTAL at https://github.com/Polymer/polymer-modulizer/pull/367/commits/e8eb2ed03867d90666e704831f0530d341e0c052. I updated the conversion functions in `PackageUrlHandler` to actually return / create what I think are correct URLs / file paths given a particular URL. This meant that I ended up removing the filtering in `PackageScanner` so that when a package is converted it doesn't ignore dependencies.

If package mode is used to convert each package within workspace mode, then I'm concerned that this will break workspace mode (or duplicate some work) by removing the filtering. Also, I'm not sure the changes I've made in `PackageUrlHandler` are really resilient to whether or not the package is being converted individually or as part of a workspace. Do original URLs pointing at dependencies start with `../` or `bower_components/`? Should convertedFilePaths point into `../@org/package` or `node_modules/@org/package`? I'm not sure that the `PackageUrlHandler` and `PackageConverter` have enough info to know the answers to those questions at the moment.

bicknellr:
Waiting on https://github.com/Polymer/polymer-modulizer/pull/372.

src/document-converter.ts
```diff
@@ -1098,6 +1113,12 @@ export class DocumentConverter {
           'shadycss/custom-style-interface.html',
           'shadycss/entrypoints/custom-style-interface.js');
     }
+    if (this.originalPackageName === 'polymer' && jsUrl === './polymer.html') {
+      jsUrl = './polymer-legacy.html';
+    } else if (jsUrl.endsWith('polymer/polymer.html')) {
```
bicknellr: I think this might need to check `originalPackageName === 'polymer'` too.

bicknellr:
I've merged in #372, so this branch now looks like it contains both given that it's based off of master. Here's the normal diff, which should show up again once #372 is merged: https://github.com/Polymer/polymer-modulizer/compare/wip-package-relative-includes...rename-polymer

FredKSchott:
Reminder (mostly for me) that this should include fixture fixes as well once #372 is merged in.

src/document-converter.ts
```diff
@@ -1098,6 +1113,12 @@ export class DocumentConverter {
           'shadycss/custom-style-interface.html',
           'shadycss/entrypoints/custom-style-interface.js');
     }
+    if (this.originalPackageName === 'polymer' && jsUrl === './polymer.html') {
```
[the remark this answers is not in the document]
bicknellr: added

bicknellr:
I'm going to update the fixtures in a separate PR so that the changes needed in this PR are cleaner.

bicknellr:
Fixtures PR: #383

FredKSchott:
Can you instead run `node lib/tools/update-fixtures.js --skip-source-update` in this PR? It will remove most of the noise and just regenerate the `expected` fixtures.

(This behavior should be more obvious, I'll create a second PR to address)

bicknellr:
@justinfagnani, mind taking another look?

bicknellr:
I need to update README.md, it has examples that would change as a result of this PR.

bicknellr:
ping @justinfagnani

src/urls/util.ts
```diff
@@ -47,7 +54,27 @@ export function getJsModuleConvertedFilePath(originalUrl: OriginalDocumentUrl):
...
+ * Converts the OriginalDocumentUrl of a file which was a script prior to
+ * conversion to its corresponding ConvertedDocumentFilePath.
+ *
+ * This function's output is affected by any renaming that might have been
+ * applied to a file, as opposed to `getJsModuleConvertedFilePath` and
+ * `getHtmlDocumentConvertedFilePath`, which consume ConvertedDocumentFilePaths
+ * which have already had renaming applied (but which haven't had their
+ * extension changed, if needed).
+ *
+ * TODO: This implementation by casting is only safe because 'polymer.html' is
```
justinfagnani: add name to todo

justinfagnani:
LGTM. Please be sure to squash.

src/urls/util.ts
bicknellr: added

PR #369, opened 2018-03-19
handle bad travis.yml

FredKSchott:
Yes, a travis.yml can exist and be empty. Because why not! 

**ex:** https://github.com/PolymerElements/app-media/blob/master/.travis.yml
**DefinitelyTyped fix here:** https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24385


azakus:
Hmm, an empty travis config would default to a `ruby` testing config, and very probably can't do the right thing.

PR #372, opened 2018-03-21
Packages now differentiate between local files and dependencies during conversion.

bicknellr:


src/urls/package-url-handler.ts
```diff
@@ -86,7 +86,7 @@ export class PackageUrlHandler implements UrlHandler {
    * package.
    */
   getOriginalPackageNameForUrl(url: OriginalDocumentUrl): string {
-    if (url.startsWith('../')) {
+    if (url.startsWith('bower_components/')) {
```
justinfagnani: this change seems to be going from detecting component-style cross-package references, to app-style - what's this for?

src/urls/package-url-handler.ts
```diff
@@ -223,14 +223,24 @@ export class PackageUrlHandler implements UrlHandler {
...
   }
 
-  packageRelativeToOriginalUrl(_originalPackageName: string, url: string):
+  packageRelativeToOriginalUrl(originalPackageName: string, url: string):
       OriginalDocumentUrl {
-    return url as OriginalDocumentUrl;
+    if (originalPackageName === this.bowerPackageName) {
+      return url as OriginalDocumentUrl;
+    } else {
+      return 'bower_components/' + originalPackageName + '/' + url as
```
justinfagnani: for component-style URls, this would start with `../`, no? Is there a reason we can treat all as app-style?

src/conversion-settings.ts
```diff
@@ -176,30 +164,19 @@ function getNamespaceNames(analysis: Analysis) {
  * incomplete user-provided options.
  */
 export function createDefaultConversionSettings(
-    analyzer: Analyzer, analysis: Analysis, options: PartialConversionSettings):
-    ConversionSettings {
+    _analyzer: Analyzer,
```
bicknellr: (from offline) Remove this unused parameter.

src/project-scanner.ts
```diff
@@ -73,21 +75,50 @@ export class ProjectScanner {
...
       return;
     }
+
+    // If the package has custom entrypoints listed in the conversion settings,
+    // use those. Otherwise, read them from the package's 'bower.json'.
+    const overriddenEntrypoints =
+        this.conversionSettings.entrypoints.get(packageName);
+    const topLevelEntrypoints = overriddenEntrypoints ?
+        overriddenEntrypoints :
+        await this.readBowerJsonEntrypoints(packageName);
```
bicknellr: (from offline) Probably cleaner as `= x && y;` instead of `= x ? x : y;`

src/test/unit/project-converter_test.ts
```diff
@@ -78,26 +78,33 @@ suite('AnalysisConverter', () => {
         bowerPackageName = 'some-package',
         packageType = 'element',
         expectedWarnings = [],
+        entrypoints = new Map(),
       } = partialOptions;
+
+      if (!entrypoints.has('some-package')) {
+        entrypoints.set('some-package', ['test.html']);
```
bicknellr: (from offline) Move this into the default for `entrypoints` above. Users should provide all of this map if they intend to provide any of it.

src/test/unit/project-converter_test.ts
```diff
@@ -78,26 +78,33 @@ suite('AnalysisConverter', () => {
...
-      conversionSettings.includes.add('test.html');
-      if (partialOptions.includes) {
-        partialOptions.includes.forEach(
-            (str) => conversionSettings.includes.add(str));
-      }
+      // conversionSettings.includes.add('test.html');
+      // if (partialOptions.includes) {
+      //   partialOptions.includes.forEach(
+      //       (str) => conversionSettings.includes.add(str));
+      // }
```
bicknellr: (from offline) remove

src/project-scanner.ts
```diff
@@ -73,21 +75,50 @@ export class ProjectScanner {
...
+    // console.log(packageName, bowerJsonPath, bowerMainFiles,
+    // topLevelEntrypoints);
```
bicknellr: (from offline) remove

src/package-scanner.ts
```diff
@@ -111,6 +114,15 @@ export class PackageScanner {
    * Scan each document in a package manually.
    */
   scanPackageManually() {
+    // Scan top-level entrypoints first, to make sure thier dependencies are
+    // properly converted to JS modules as well.
```
bicknellr: (from offline) Pull the logic that switches the scan result type based on `topLevelEntrypoints` out of `scanDocument` and only do it here.

src/urls/package-url-handler.ts
bicknellr: My understanding is that a `PackageUrlHandler` is only used when converting in project mode, so dependencies of the main package will always appear under `bower_components/`. @FredKSchott, please correct me if I'm wrong.

src/urls/package-url-handler.ts
bicknellr: I think my [comment above](https://github.com/Polymer/polymer-modulizer/pull/372/files#r176244380) also applies here: `PackageUrlHandler` is only used when converting in project mode, so all files that are part of dependencies will be in `bower_components/`.

src/urls/package-url-handler.ts
bicknellr: Following from that, I *think* that dependencies in workspace mode are all converted using a single `WorkspaceUrlHandler`, rather than individual `PackageUrlHandler`s.

src/test/fixtures/fixtures_test.ts
```diff
@@ -127,6 +127,8 @@ suite('Fixtures', () => {
         const output = await runFixture(
             fixtureSourceDir, fixtureResultDir, fixtureTestConfig);
 
+        // console.log(output.stdout);
+
```
bicknellr: (from offline) remove

src/urls/package-url-handler.ts
FredKSchott: - this has nothing to do with element vs. application. It has to do with the URL format the analyzer gives us (pre 3.0 it gave us `../`, post 3.0 it gives us 'bower_components/')

src/project-scanner.ts
bicknellr: sorry `x && y` should be `x || y`

src/conversion-settings.ts
bicknellr: removed

src/package-scanner.ts
bicknellr: done

src/project-scanner.ts
bicknellr: done

src/project-scanner.ts
bicknellr: done

src/test/fixtures/fixtures_test.ts
bicknellr: done

src/test/unit/project-converter_test.ts
bicknellr: done

src/test/unit/project-converter_test.ts
bicknellr: done

src/project-scanner.ts
```diff
@@ -73,21 +75,47 @@ export class ProjectScanner {
     return packageScanner.getConversionManifest();
   }
 
+  /**
+   * Reads the entrypoints for a package from the `main` field in the package's
+   * 'bower.json' file.
+   */
+  private async readBowerJsonEntrypoints(packageName: string) {
+    const bowerJsonPath =
+        this.urlHandler.packageRelativeToOriginalUrl(packageName, 'bower.json');
```
bicknellr: This should be `path.join(this.urlHandler.getPackageDir(packageName), 'bower.json')`.

src/project-scanner.ts
```diff
@@ -73,21 +75,47 @@ export class ProjectScanner {
...
+    const bowerJson = await fse.readJSON(bowerJsonPath) as Partial<BowerConfig>;
```
bicknellr: `fse.readJSON` rejects when the file doesn't exist, so this needs to be wrapped in a try/catch which sets `bowerMainFiles` to `[]` with a console warning?

bicknellr: fixed

src/project-scanner.ts
bicknellr: Fixed, along with fixing `getPackageDir`.

fixtures/packages/polymer/expected/test/unit/shady-unscoped-style-import-css.js
```diff
@@ -1 +0,0 @@
-
```
FredKSchott: Note for other reviewers: This change is due to how we previously incorrectly marked this file for conversion. We now correctly recognize that this file is imported as CSS.

src/conversion-settings.ts
```diff
@@ -31,32 +33,16 @@ export interface ConversionSettings {
   readonly namespaces: Set<string>;
 
   /**
-   * Files to exclude from conversion (ie lib/utils/boot.html).
+   * A map from original package names to entrypoints to be used when converting
+   * each listed package. Setting an entry in this map completely replaces the
+   * default entrypoints for that package.
    */
-  readonly excludes: Set<string>;
+  readonly entrypoints: Map<string, OriginalDocumentUrl[]>;
```
FredKSchott: since this is a map, I think it would benefit from a more descriptive name. How about `packageEntrypoints` to indicate that this is a map of package-based entrypoints?

src/conversion-settings.ts
```diff
@@ -105,6 +91,8 @@ export interface PartialConversionSettings {
    */
   readonly namespaces?: Iterable<string>;
 
+  readonly entrypoints?: Map<string, OriginalDocumentUrl[]>;
```
FredKSchott: ditto

src/conversion-settings.ts
```diff
@@ -176,30 +164,18 @@ function getNamespaceNames(analysis: Analysis) {
...
 export function createDefaultConversionSettings(
-    analyzer: Analyzer, analysis: Analysis, options: PartialConversionSettings):
-    ConversionSettings {
+    analysis: Analysis,
+    options: PartialConversionSettings): ConversionSettings {
   // Configure "namespaces":
   const namespaces =
       new Set(getNamespaceNames(analysis).concat(options.namespaces || []));
 
+  const entrypoints = options.entrypoints || new Map();
```
FredKSchott: add `// Configure entrypoints` comment

src/conversion-settings.ts
```diff
@@ -176,30 +164,18 @@ function getNamespaceNames(analysis: Analysis) {
...
+
   // Configure "excludes":
   const excludes = new Set(
       [...(options.excludes || []), 'neon-animation/web-animations.html']);
 
-  // Configure "includes":
```
FredKSchott: 👏 👏 👏

src/package-scanner.ts
```diff
@@ -114,8 +117,17 @@ export class PackageScanner {
...
     for (const document of this.getPackageHtmlDocuments()) {
-      this.scanDocument(document);
+      if (this.topLevelEntrypoints.has(
+              this.urlHandler.getDocumentUrl(document))) {
+        this.scanDocument(document, 'js-module');
+      }
+    }
+    // Scan all other documents, to be converted as top-level HTML files.
+    for (const document of this.getPackageHtmlDocuments()) {
```
FredKSchott: Right now this iterates over all documents and tries to scan each one to an html-document, which sounds wrong. This PR does a good job of moving off the side-effects of scanDocument, could we go even further and add a `shouldScanDocument()` check here as well?

src/project-scanner.ts
```diff
@@ -73,21 +76,59 @@ export class ProjectScanner {
     return packageScanner.getConversionManifest();
   }
 
+  /**
+   * Reads the entrypoints for a package from the `main` field in the package's
+   * 'bower.json' file.
+   */
+  private async readBowerJsonEntrypoints(packageName: string) {
```
FredKSchott: `packageName` -> `bowerPackageName`

src/project-scanner.ts
```diff
@@ -73,21 +76,59 @@ export class ProjectScanner {
...
+
+    let bowerMainFiles = bowerJson ? bowerJson.main : [];
+    if (typeof bowerMainFiles === 'string') {
+      bowerMainFiles = [bowerMainFiles];
+    } else if (bowerMainFiles === undefined) {
+      bowerMainFiles = [];
+    }
+
+    return bowerMainFiles.map((relativeOriginalUrl) => {
+      return this.urlHandler.packageRelativeToOriginalUrl(
```
FredKSchott: we'll usually remove the `return` statement when we use a single-line callback.

```
return bowerMainFiles.map((relativeOriginalUrl) => 
  this.urlHandler.packageRelativeToOriginalUrl(packageName, relativeOriginalUrl));
```

src/project-scanner.ts
```diff
@@ -73,21 +76,59 @@ export class ProjectScanner {
...
+          packageName, relativeOriginalUrl);
+    });
+  }
+
   /**
    * Scan a document and any of its dependency packages for their new interface.
    */
-  async scanPackage(matchPackageName: string, forceScan = false) {
-    if (this.scannedPackages.has(matchPackageName)) {
+  async scanPackage(packageName: string, forceScan = false) {
```
FredKSchott: `bowerPackageName` (now I'm remembering, maybe you already do this in #367. If so, fine to ignore here).

src/test/unit/project-converter_test.ts
```diff
@@ -74,7 +74,7 @@ suite('AnalysisConverter', () => {
       packageType: PackageType;
       npmImportStyle: NpmImportStyle;
       expectedWarnings: string[];
-      includes: string[];
+      entrypoints: Map<string, OriginalDocumentUrl[]>;
```
FredKSchott: I have one last idea to try to get rid of all the `OriginalDocumentUrl` casting: can we not extend `PartialConversionSettings` here, and then set `entrypoints` to `Map<string, string[]>`? We should still get proper type warning if we set bad properties on `partialSettings`.

src/conversion-settings.ts
bicknellr: Updated to `packageEntrypoints`.

src/conversion-settings.ts
bicknellr: -> `packageEntrypoints`

src/conversion-settings.ts
bicknellr: Added.

src/project-scanner.ts
bicknellr: fixed

src/project-scanner.ts
bicknellr: fixed

src/project-scanner.ts
bicknellr: fixed

src/package-scanner.ts
bicknellr: added

bicknellr:
NP, thanks for your help on this PR! Merging once the last CI run has passed.

PR #378, opened 2018-03-25
this.x = assignments in namespace functions make properties mutable

### LINKED ISSUE #374: Namespace properties converted as const, should be let
See https://github.com/PolymerElements/iron-overlay-behavior/blob/__auto_generated_3.0_preview/iron-scroll-manager.js#L101

These properties are assigned elsewhere in the file, which throws.

justinfagnani:
Fixes #374

### DIFF
--- a/src/passes/rewrite-namespace-exports.ts
+++ b/src/passes/rewrite-namespace-exports.ts
@@ -139,6 +139,10 @@ class RewriteNamespaceExportsPass {
     const fullyQualifiedName = memberPath.join('.');
     this.namespaceNames.add(fullyQualifiedName);
 
+    // Note: in other place in this file where we decide between let and const
+    // exports, we check if the "this." name is mutable as well, but in this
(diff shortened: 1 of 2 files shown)


e111077:
I think your link is pointing to the wrong bug?

e111077:
This _looks_ right to me but I feel uncomfortable being the only LGTM in this PR.

justinfagnani:
Oops. should be #374

src/passes/rewrite-namespace-exports.ts
```diff
@@ -139,6 +139,10 @@ class RewriteNamespaceExportsPass {
     const fullyQualifiedName = memberPath.join('.');
     this.namespaceNames.add(fullyQualifiedName);
 
+    // Note: in other place in this file where we decide between let and const
```
FredKSchott: 👍

PR #381, opened 2018-03-26
update dependency map for pre.12, unmapped 3rd party dependencies

FredKSchott:
- After each release, we update the dependency map to the new released version tag
- We've decided to remove the leading `^` from our own dependencies going forward, so that we can test breaking changes under the `next` tag before sharing them with our users.
- This PR also adds some previously unmapped dependencies

### DIFF
--- a/dependency-map.json
+++ b/dependency-map.json
@@ -1,235 +1,235 @@
 {
   "app-layout": {
     "npm": "@polymer/app-layout",
-    "semver": "^3.0.0-pre.10"
+    "semver": "3.0.0-pre.12"
   },
   "app-elements": {
     "npm": "@polymer/app-elements",
-    "semver": "^3.0.0-pre.10"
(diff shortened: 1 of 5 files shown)


dependency-map.json
```diff
@@ -497,7 +497,7 @@
   },
   "polymer": {
     "npm": "@polymer/polymer",
-    "semver": "^3.0.0-pre.10"
+    "semver": "3.0.0-pre.12"
```
rictic: Was removing the leading ^ intentional here?

rictic:
Looks like some goldens need updating as well

dependency-map.json
FredKSchott: yes, OP has been updated with the rational for the change

FredKSchott:
@rictic you're right! Pushed fixes and re-running tests...

PR #382, opened 2018-03-26
Separate out document conversion from scanning.

rictic:
Makes the document scanning vs converting process a bit easier to follow.

no-op change.

### DIFF
--- a/src/document-processor.ts
+++ b/src/document-processor.ts
@@ -1,0 +1,400 @@
+/**
+ * @license
+ * Copyright (c) 2018 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as dom5 from 'dom5';
+import {Program} from 'estree';
+import * as jsc from 'jscodeshift';
+import * as parse5 from 'parse5';
+import {Document, Import, isPositionInsideRange, Severity, Warning} from 'polymer-analyzer';
+import * as recast from 'recast';
+
+import {ConversionSettings} from './conversion-settings';
+import {attachCommentsToFirstStatement, canDomModuleBeInlined, createDomNodeInsertStatements, filterClone, getCommentsBetween, getNodePathInProgram, insertStatementsIntoProgramBody, serializeNodeToTemplateLiteral} from './document-util';
+import {NamespaceMemberToExport} from './js-module';
+import {removeNamespaceInitializers} from './passes/remove-namespace-initializers';
+import {removeToplevelUseStrict} from './passes/remove-toplevel-use-strict';
+import {removeUnnecessaryEventListeners} from './passes/remove-unnecessary-waits';
+import {removeWrappingIIFEs} from './passes/remove-wrapping-iife';
+import {rewriteToplevelThis} from './passes/rewrite-toplevel-this';
+import {ConvertedDocumentFilePath, ConvertedDocumentUrl, OriginalDocumentUrl} from './urls/types';
+import {UrlHandler} from './urls/url-handler';
+import {isOriginalDocumentUrlFormat} from './urls/util';
+import {replaceHtmlExtensionIfFound} from './urls/util';
+
+/**
+ * Keep a set of elements to ignore when Recreating HTML contents by adding
+ * code to the top of a program.
+ */
+const generatedElementBlacklist = new Set<string|undefined>([
+  'base',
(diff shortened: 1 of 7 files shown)


src/document-processor.ts
```diff
@@ -0,0 +1,400 @@
...
+export interface HtmlDocumentScanResult {
+  type: 'html-document';
+  originalUrl: OriginalDocumentUrl;
+  convertedUrl: ConvertedDocumentUrl;
+  convertedFilePath: ConvertedDocumentFilePath;
+}
+
+
+/**
+ * Converts a Document from Bower to NPM. This supports converting HTML files
```
FredKSchott: update docs. Would love to see a summary of how this is expected to be used (directly? as a base class?)

src/document-processor.ts
```diff
@@ -0,0 +1,400 @@
...
+  originalUrl: OriginalDocumentUrl;
+  convertedUrl: undefined;
+  convertedFilePath: undefined;
+}
+
+/**
+ * Contains information that an existing file should be converted as a top-level
+ * HTML file (and not as a new JS module).
+ */
+export interface HtmlDocumentScanResult {
```
FredKSchott: move these types to `document-scanner.ts`

src/document-processor.ts
```diff
@@ -0,0 +1,400 @@
...
+      removeToplevelUseStrict(scriptProgram);
+      // We need to inline templates on a per-script basis, otherwise we run
+      // into trouble matching up analyzer AST nodes with our own.
+      const localClaimedDomModules =
+          this.inlineTemplates(scriptProgram, scriptDocument);
+      for (const claimedDomModule of localClaimedDomModules) {
+        claimedDomModules.add(claimedDomModule);
+      }
+      if (this.conversionSettings.addImportPath) {
+        this.addImportPathsToElements(scriptProgram, scriptDocument);
```
FredKSchott: This feels like a "conversion" concern, can it be moved to DocumentConverter?

src/document-processor.ts
```diff
@@ -0,0 +1,400 @@
...
+        this.document.parsedDocument.ast, prevScriptNode, undefined);
+    const maybeCommentStatement =
+        attachCommentsToFirstStatement(trailingComments, []);
+    combinedToplevelStatements.push(...maybeCommentStatement);
+    const program = jsc.program(combinedToplevelStatements);
+    removeUnnecessaryEventListeners(program);
+    removeWrappingIIFEs(program);
+
+    this.insertCodeToGenerateHtmlElements(program, claimedDomModules);
+    removeNamespaceInitializers(program, this.conversionSettings.namespaces);
```
FredKSchott: Same here, moved to DocumentConverter?

src/document-processor.ts
```diff
@@ -0,0 +1,400 @@
...
+  protected readonly convertedHtmlScripts: ReadonlySet<Import>;
+
+  constructor(
+      document: Document, urlHandler: UrlHandler,
+      conversionSettings: ConversionSettings) {
+    this.conversionSettings = conversionSettings;
+    this.urlHandler = urlHandler;
+    this.document = document;
+    this.originalUrl = urlHandler.getDocumentUrl(document);
+    this.convertedUrl = this.convertDocumentUrl(this.originalUrl);
```
FredKSchott: some background here: The expected behavior of this property is pretty confusing. It is always the "html -> js module converted URL", even if the file is eventually converted as a top-level HTML document. This flaw is even exposed in the return result.

This was only just realized a few weeks ago, since `convertedUrl` is mainly used to create relative URLs, and `convertedFilePath` (which is used to write the conversions to disk) does have the correct file extension.

This doesn't need to be tackled in this PR, but I thought I'd mention it. I'll create a separate issue to track.

src/document-processor.ts
rictic: done

src/document-processor.ts
rictic: Done, PTAL

src/document-processor.ts
rictic: Noted on the field.

src/document-processor.ts
rictic: I don't want to move this mutation out of here yet, my goal here was for this to be a simple noop code split. Added a comment+TODO

src/document-processor.ts
rictic: Same

src/document-processor.ts
```diff
@@ -0,0 +1,365 @@
...
+    const isInternalImport =
+        this.urlHandler.isImportInternal(this.convertedUrl, newScriptUrl);
+    return isInternalImport && !isModuleImport;
+  }
+
+  /**
+   * Creates a single program from all the JavaScript in the current document.
+   * The standard program result can be used for either scanning or conversion.
+   *
+   * TODO: this does a lot of mutation of the program. Could we only do that
```
FredKSchott: I talked about this with Russel & Justin a bit today, but +1. It would be great if scanning didn't modify the formatted/processed `program` at all, and instead returned data that would then inform the scan results & different conversion steps. 

That way scanning passes could be decoupled/not order-dependent, and conversion passes could be even simpler and more easily configured.

PR #385, opened 2018-03-26
Preserve comments in rewrite namespace exports pass.

rictic:
Analyzing and linting polymer 3 sources was having lots of issues due to load-bearing-comments being missing.

### DIFF
--- a/src/passes/rewrite-namespace-exports.ts
+++ b/src/passes/rewrite-namespace-exports.ts
@@ -433,17 +433,38 @@ function getMutableNames(program: estree.Program): Set<string> {
   return mutable;
 }
 
+const jsdocToRemove = /@(namespace|memberof)/;
 function replacePreservingComments(
     nodePath: NodePath, replacement: estree.Node) {
   const comments = getComments(nodePath);
   nodePath.replace(replacement);
+  const commentsToRemove = new Set<estree.Comment>();
   for (const comment of comments) {
+    if (!jsdocToRemove.test(comment.value)) {
+      continue;
+    }
     // Filter out namespace and memberof comments, which no longer make sense.
     const lines = comment.value.split('\n');
     comment.value =
-        lines.filter((l) => !/@(namespace|memberof)/.test(l)).join('\n');
+        lines.filter((l) => !jsdocToRemove.test(l))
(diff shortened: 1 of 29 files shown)


src/passes/rewrite-namespace-exports.ts
```diff
@@ -121,9 +121,12 @@ class RewriteNamespaceExportsPass {
         correctedNamespaceName = 'Polymer';
         name = 'Polymer';
       }
-      nodePath.replace(jsc.exportNamedDeclaration(jsc.variableDeclaration(
-          this.mutableNames.has(correctedNamespaceName) ? 'let' : 'const',
-          [jsc.variableDeclarator(jsc.identifier(name), exportedExpression)])));
+      replacePreservingComments(
```
FredKSchott: so many lines! Can you split these up a bit with a few variable declarations?

src/test/unit/project-converter_test.ts
```diff
@@ -910,10 +910,14 @@ export function increment() {
       });
       assertSources(await convert(), {
         'test.js': `
+/**
```
FredKSchott: if comment is empty like this, we should continue to remove it

src/test/unit/project-converter_test.ts
```diff
@@ -2951,6 +2963,52 @@ console.log('second script');
...
+             */
+            Polymer.MyMixinFunction = function() {};
+          </script>
+        `
+      });
+
+      assertSources(await convert(), {
+        'test.js': `
+/**
+             * This is a long and important comment.
```
FredKSchott: can we fix this formatting?

src/test/unit/project-converter_test.ts
rictic: Seems reasonable.

src/test/unit/project-converter_test.ts
rictic: The recast API is not well documented. I poked around in the debugger but didn't find anything that did exactly what we want. Hacked something together that seems to do all right though.

rictic:
Comments addressed, PTAL

rictic:
Thanks!

PR #388, opened 2018-03-27
Remove `include` CLI flag.

bicknellr:
This should have happened in #372. Doesn't seem like there was any documentation for `includes` other than comments in the code, so I haven't updated anything surrounding that.

### DIFF
--- a/src/convert-package.ts
+++ b/src/convert-package.ts
@@ -14,7 +14,7 @@
 
 import * as fse from 'fs-extra';
 import * as path from 'path';
-import {Analysis, Analyzer, FSUrlLoader, InMemoryOverlayUrlLoader, PackageUrlResolver, ResolvedUrl} from 'polymer-analyzer';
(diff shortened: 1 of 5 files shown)


src/convert-package.ts
```diff
@@ -60,16 +60,6 @@ async function setupOutDir(outDir: string, clean = false) {
...
-/**
- * Create the default conversion settings, adding any "main" files from the
- * current package's bower.json maifest to the "includes" set.
- */
-function getConversionSettings(
-    analysis: Analysis, options: PackageConversionSettings) {
-  const conversionSettings = createDefaultConversionSettings(analysis, options);
-  return conversionSettings;
-}
-
```
bicknellr: I was going to update the comment, then I realized that this function does nothing so I just removed it...

FredKSchott: Ha! nice

bicknellr:
Maybe, I should update CHANGELOG.md here too... I could leave a small tip to using `main` there.

PR #390, opened 2018-03-29
update fixtures

bicknellr:
This is mostly about getting an update to Polymer's bower.json which adds iron-component-page as a dev dependency:
Polymer commit: https://github.com/Polymer/polymer/commit/d93dd1ceeb76291496204634e3df96dc959c69d5
Local blob: https://github.com/Polymer/polymer-modulizer/blob/b9a949c8dd8e46da874faa406cb761b922ad5102/fixtures/packages/polymer/source/bower.json


aomarks:
It's kinda crazy how much code we check into this repo. We could just store the SHAs and check them out upon install...

bicknellr:
Updated the test timeout, WDYT?

bicknellr:
Having a bunch of individual submodules might be a bit messy but maybe if all of the fixtures were moved to a separate repo and that was made a single submodule here, things might be a bit cleaner.

bicknellr:
(cc @rictic: timeout update)

PR #392, opened 2018-04-04
Update analyzer to ^3.0.0-pre.20

bicknellr:
The new version of analyzer produces `Import`s where `document` may be undefined when it previously wouldn't have produced them at all, so this PR just throws in all of the places where `document` is used if `document` is undefined. Generally this is a good thing because it points out references to files that don't exist but it also means that workspace conversion is broken for the current set of repos listed in `docs/polymer-modulize-repos.txt` because some of these repos have incorrect references. I've been working on getting fixes into them but app-layout is one I'm not particularly sure what to do with: it contains a bunch of packages in [`/templates`](https://github.com/PolymerElements/app-layout/tree/0e237cdab14f2368f4f7fd3aa06af4ee5571c0ca/templates) which expect you to install things with bower and the files in those directories reach directly into `bower_components` to get them.


bicknellr:
Updated so that at the places where imports are retrieved from analyzer, imports without a document are filtered out with a console warning.

src/document-converter.ts
```diff
@@ -235,10 +248,18 @@ export class DocumentConverter extends DocumentProcessor {
     // itself.
     for (const scriptImport of this.document.getFeatures(
              {kind: 'html-script'})) {
+      if (scriptImport.document === undefined) {
+        console.warn(
+            `${this.originalPackageName} ${this.originalUrl}: ` +
+            `The script referenced using URL '${scriptImport.originalUrl}' ` +
+            `not be loaded and was ignored.`);
```
bicknellr: grammar: missing 'could'

src/package-scanner.ts
```diff
@@ -270,7 +271,22 @@ export class PackageScanner {
...
+    const htmlImports = DocumentConverter.getAllHtmlImports(document);
+    const htmlImportsWithDocuments =
+        htmlImports.filter((htmlImport: Import) => {
+          const document = htmlImport.document;
+          const hasDocument = document !== undefined;
+          if (!hasDocument) {
+            console.warn(
+                `${packageName} ${documentUrl}: The document referenced ` +
+                `using URL '${
+                    htmlImport.originalUrl}' could not be loaded and ` +
```
bicknellr: bad formatting

bicknellr: fixed

src/document-converter.ts
bicknellr: fixed

src/import-with-document.ts
```diff
@@ -0,0 +1,5 @@
+import {Document, Import} from 'polymer-analyzer';
+
+export interface ImportWithDocument extends Import {
+  readonly document: Document;
+}
```
rictic: Can also define a function that will tell you if you have an ImportWithDocument:

```
function importHasDocument(import_: Import): import_ is ImportWithDocument {
  return document !== undefined;
}
```

Then we can use this as a type safe alternative to casting with `as` (which should be avoided as much as possible).

bicknellr: Very cool; I was trying out Flow recently and was trying to figure out how to write a type checking function but couldn't get past the fact that the narrowing only applied within the function. :/ It's neat that TypeScript has an explicit way to do this. I've updated the PR with an `isImportWithDocument` function.

bicknellr:
Is it better to warn [like this](https://github.com/Polymer/polymer-modulizer/blob/bfcb8520fdc370914b22dd69707b3ff57f9e340a/src/document-processor.ts#L230)?

bicknellr:
Tried it - yes it definitely is better.

bicknellr:
I think this is ready for review again.

src/import-with-document.ts
rictic: That flow doc is neat!

I didn't realize that they did type checking of function parameters by looking at everywhere it's called. That seems like too much inference to me, but that might just be change aversion talking.

I really wish that `Array#filter` would flow through predicate functions in typescript.

src/document-converter.ts
```diff
@@ -219,10 +219,14 @@ export class DocumentConverter extends DocumentProcessor {
     const filteredImports = [];
     for (const import_ of DocumentConverter.getAllHtmlImports(this.document)) {
       if (!isImportWithDocument(import_)) {
-        console.warn(
-            `${this.originalPackageName} ${this.originalUrl}: The document ` +
-            `referenced using URL '${import_.originalUrl}' could not be ` +
-            `loaded and was ignored.`);
+        console.warn(new Warning({
+          code: 'import-ignored',
+          message: `The import referenced by URL '${import_.originalUrl}' ` +
```
rictic: No need to include info that's present in the source code, it'll be logged with a squiggle under the original source code. So these messages could just be:

```
message: `Import could not be loaded and will be ignored.`,
```

src/import-with-document.ts
bicknellr: > I really wish that `Array#filter` would flow through predicate functions in typescript.

Same; this is actually why I changed `DocumentConverter#getHtmlImports` to a for loop in this PR.

src/document-converter.ts
bicknellr: Ok, I'll update these messages and merge.

PR #393, opened 2018-04-05
Package type

### LINKED ISSUE #206: Problem modulizing polymer/shop
`polymer/shop` has a lazy resources html import: https://github.com/Polymer/shop/blob/master/src/lazy-resources.html

When I ran `modulizer --out . --force` on `shop`, it did this: https://github.com/Polymer/shop/blob/8b7665099fe5094f39fe73ad825f5d00fc6471be/src/lazy-resources.html
- all the imports inside `lazy-resources` get converted to `script type=module`
- the file itself isn't renamed, and then you can't lazy import it with `import('./lazy-resources.html')` without a "wrong mimetype" warning

I think what should happen instead is this: https://github.com/Polymer/shop/blob/3.0/src/lazy-resources.js
- all the imports inside it get converted to `import foo` statements
- the file gets renamed to `lazy-resources.js`

/cc @frankiefu

e111077:
Fixes #206

We seem to have been ignoring the type of project we are trying to modulize. This affects whether it's a relative path out of the package or into `node_modules`. This can be set via the `--package-type` flag which defaults to `element`.

I had it as a mandatory flag where it would ask you to choose between element and application but I changed my mind because I personally didn't like having to type it in every time. LMK if I should bring it back.

CC @notwaldorf

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -13,7 +13,7 @@
  */
 
 import * as commandLineArgs from 'command-line-args';
-import {NpmImportStyle} from '../conversion-settings';
+import {NpmImportStyle, PackageType} from '../conversion-settings';
 
 import runPackageCommand from './command-package';
 import runWorkspaceCommand from './command-workspace';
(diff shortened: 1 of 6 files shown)


src/cli/index.ts
```diff
@@ -248,5 +257,12 @@ installation.
         `Supported styles: "name", "path".`);
   }
 
+  const packageType = options['package-type'];
+  if (packageType !== 'element' && packageType !== 'application') {
```
usergenic: If you had a `packageTypes` variable you could use it all over the place and do `!packageTypes.includes(packageType)` and like `Supported types: "${packageTypes.join('", "')}"` and stuff but this is fine.

PR #401, opened 2018-04-15
Comments comments comments

rictic:
Depends on https://github.com/Polymer/tools/pull/150 and will fail until that lands.

* HTML comments that come after all scripts will be inserted at the end of the output JS.
* HTML comments that come before all scripts will be inserted at the very very beginning of the output JS, even before imports.
* If an HTML comment is just an `@license` comment, it doesn't need a `FIXME(polymer-modulizer)` comment when we stick it at the beginning of the output JS. That's where it's supposed to go.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1807,9 +1807,9 @@
       }
     },
     "clone": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/clone/-/clone-2.1.1.tgz",
-      "integrity": "sha1-0hfR6WERjjrJpLi7oyhVU79kfNs="
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/clone/-/clone-2.1.2.tgz",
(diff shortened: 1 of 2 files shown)


src/document-processor.ts
```diff
@@ -107,6 +110,12 @@ export abstract class DocumentProcessor {
     const convertedHtmlScripts = new Set<ImportWithDocument>();
     const claimedDomModules = new Set<parse5.ASTNode>();
     let prevScriptNode: parse5.ASTNode|undefined = undefined;
+    /**
+     * Inserting leading comments is surprisingly tricky, because they must
+     * be attached to a node, but very few nodes can come before an import,
+     * and in general we don't want to insert fake nodes, so instead we
+     * pass these up to be added onto the first node in the
```
aomarks: in the...

rictic: Fixed in 7740afe

PR #411, opened 2018-04-30
Avoid using nested templates when setting innerHTML

### LINKED ISSUE #410: IE11: Converting HTML to innerHTML that's written to main doc does not work with <template> polyfill
Due to a [limitation](https://github.com/webcomponents/template/issues/40) in the Template polyfill, nested templates are not upgraded when `innerHTML` is set on a non-template element.

This means that when modulizer converts HTML to a string that's set as `innerHTML` on an element and then attached to the main document, any <templates> will not be upgraded and therefore not work on browsers (like IE11) that do not have <template>.

To fix this, we can change the pattern used to this:

```
const documentContainer = document.createElement('template');
documentContainer.innerHTML = `<div style="display: none;">HTML CONTENT</div`;
document.body.appendChild(documentContainer.content);
```

web-padawan:
Fixes #410 by applying the change from Polymer/polymer#5210

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-modulizer/411)

### DIFF
--- a/src/document-util.ts
+++ b/src/document-util.ts
@@ -540,7 +540,7 @@ export function createDomNodeInsertStatements(
           jsc.callExpression(
               jsc.memberExpression(
                   jsc.identifier('document'), jsc.identifier('createElement')),
-              [jsc.literal('div')]))]);
+              [jsc.literal('template')]))]);
   const setDocumentContainerStatement =
       jsc.expressionStatement(jsc.assignmentExpression(
           '=',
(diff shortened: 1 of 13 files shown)


e111077:
You on on point! We were just about to tackle this issue only to find that you've fixed it already with our proposed solution!

src/document-util.ts
```diff
@@ -556,23 +556,25 @@ export function createDomNodeInsertStatements(
               jsc.memberExpression(
                   jsc.identifier('document'), jsc.identifier('body')),
               jsc.identifier('appendChild')),
-          [jsc.identifier(varName)]))
+          [jsc.memberExpression(
+              jsc.identifier(varName), jsc.identifier('content'))]))
     ];
   }
-  const setDisplayNoneStatment = jsc.expressionStatement(jsc.callExpression(
+  const setDisplayNoneStatement = jsc.expressionStatement(jsc.callExpression(
```
e111077: 🍻

src/document-util.ts
```diff
@@ -540,7 +540,7 @@ export function createDomNodeInsertStatements(
           jsc.callExpression(
               jsc.memberExpression(
                   jsc.identifier('document'), jsc.identifier('createElement')),
-              [jsc.literal('div')]))]);
+              [jsc.literal('template')]))]);
```
e111077: 💯

justinfagnani:
Thanks @web-padawan !

Issue #345, opened 2018-02-28
Namespaces are being omitted


e111077:
When a namespace is defined, it seems as if the modulizer is ignoring it. e.g.

```js
/**
 * @namespace
 * @memberof Polymer
 */
Polymer.MyNamespace {
  /**
   * @memberof Polymer.MyNamespace
   */
  myMethod() { return somFn.bind(this); }
}
```

is converted to
```js
export function myMethod() { return someFn.bind(Polymer.MyNamespace); }
// but Polymer is undefined and so is MyNamespace
```

see [PolymerElements/iron-overlay-behavior/iron-scroll-manager.html _lockScrollInteractions](https://github.com/PolymerElements/iron-overlay-behavior/blob/__auto_generated_3.0_preview/iron-scroll-manager.js#L162)

[before](https://github.com/PolymerElements/iron-overlay-behavior/blob/b7afaae994d527c8582d6ae8a2694e81c571e092/iron-scroll-manager.html#L224)

[after](https://github.com/PolymerElements/iron-overlay-behavior/blob/__auto_generated_3.0_preview/iron-scroll-manager.js#L164)

justinfagnani:
This is a case that our `this` rewriting for namespaces isn't catching.

The specific lines being transformed incorrectly are:
- input: https://github.com/PolymerElements/paper-checkbox/blob/master/paper-checkbox.html#L239
- output: https://github.com/PolymerElements/iron-overlay-behavior/blob/__auto_generated_3.0_preview/iron-scroll-manager.js#L164

A simplified version is:

```js
Polymer.MyNamespace {
  methodA() {...}

  methodB() {
    return this.methodA.bind(this);
  }
}
```

I'd rather fix this by hand after conversion, but if the goal is to keep 2.x source as the source-of-truch, then we'll have to extend the heuristic to recognize `this.<function>.bind(this)` so we can output:

```js
export function methodA() {...}

export function methodB() {
  return methodA;
}
```

justinfagnani:
cc @FredKSchott @kevinpschaaf and @graynorton 

I'm not sure how many more issues like this we have to get to 100% automation on our elements, but we're in the long tail for sure, IMO.

e111077:
spoken offline, we also suggested fixing this issue by making modulizer substitute `undefined` for `this` when used in a function

bicknellr:
Is this closable after #345 or is this different?

PR #416, opened 2018-05-03
Rewrite NS object references that don't reference members to undefined.

### LINKED ISSUE #415: References to namespace objects should become `undefined` when referencing the namespace itself.
References to namespace objects which do not start a member expression should be rewritten to `undefined`, rather than the identifier used to represent the namespace object.

Input:

```html
<script>
/** @namespace*/
window.NS = {
  f: function() {
    console.log('f', this);
    this.g();
  },
  g: function() { console.log('g') },
};
</script>
```

Actual Output:

```javascript
export function f() {
  console.log('f', NS);
  g();
}

export function g() { console.log('g'); }
```

Expected Output:

```javascript
export function f() {
  console.log('f', undefined);
  g();
}

export function g() { console.log('g'); }
```

---
https://github.com/Polymer/polymer-modulizer/blob/15f925f3ab97268b4c2f9ce61e1cf55f321bd278/src/passes/rewrite-namespace-this-references.ts#L27-L30

bicknellr:
Fixes #415.

### DIFF
--- a/src/passes/rewrite-namespace-this-references.ts
+++ b/src/passes/rewrite-namespace-this-references.ts
@@ -25,7 +25,12 @@ function rewriteSingleScopeThisReferences(
     blockStatement: estree.BlockStatement, namespaceReference: string) {
   astTypes.visit(blockStatement, {
     visitThisExpression(path: NodePath<estree.ThisExpression>) {
-      path.replace(jsc.identifier(namespaceReference));
+      const parent = path.parent;
+      if (parent && parent.node.type !== 'MemberExpression') {
+        path.replace(jsc.unaryExpression('void', jsc.literal(0)));
+      } else {
+        path.replace(jsc.identifier(namespaceReference));
+      }
       return false;
     },
 
--- a/src/passes/rewrite-namespace-this-references.ts
+++ b/src/passes/rewrite-namespace-this-references.ts
@@ -19,7 +19,9 @@ import * as jsc from 'jscodeshift';
 
 /**
  * Rewrite `this` references to the explicit namespaceReference identifier
- * within a single BlockStatement. Don't traverse deeper into new scopes.
+ * within a single BlockStatement if they are the start of a member expression,
+ * otherwise they are rewritten to `undefined`. Don't traverse deeper into new
+ * scopes.
  */
 function rewriteSingleScopeThisReferences(
     blockStatement: estree.BlockStatement, namespaceReference: string) {
@@ -29,8 +31,8 @@ function rewriteSingleScopeThisReferences(
       if (parent && parent.node.type !== 'MemberExpression') {
         // When a namespace object is itself referenced with `this` but isn't
         // used to reference a member of the namespace, rewrite the `this` to
-        // `void 0`:
-        path.replace(jsc.unaryExpression('void', jsc.literal(0)));
+        // `undefined`:
+        path.replace(jsc.identifier('undefined'));
       } else {
         path.replace(jsc.identifier(namespaceReference));
       }
--- a/src/passes/rewrite-namespace-this-references.ts
+++ b/src/passes/rewrite-namespace-this-references.ts
@@ -27,6 +27,9 @@ function rewriteSingleScopeThisReferences(
     visitThisExpression(path: NodePath<estree.ThisExpression>) {
       const parent = path.parent;
       if (parent && parent.node.type !== 'MemberExpression') {
+        // When a namespace object is itself referenced with `this` but isn't
+        // used to reference a member of the namespace, rewrite the `this` to
+        // `void 0`:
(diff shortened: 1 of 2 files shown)


bicknellr:
Still checking this with the current elements...

src/passes/rewrite-namespace-this-references.ts
```diff
@@ -25,7 +25,15 @@ function rewriteSingleScopeThisReferences(
     blockStatement: estree.BlockStatement, namespaceReference: string) {
   astTypes.visit(blockStatement, {
     visitThisExpression(path: NodePath<estree.ThisExpression>) {
-      path.replace(jsc.identifier(namespaceReference));
+      const parent = path.parent;
+      if (parent && parent.node.type !== 'MemberExpression') {
+        // When a namespace object is itself referenced with `this` but isn't
+        // used to reference a member of the namespace, rewrite the `this` to
+        // `void 0`:
+        path.replace(jsc.unaryExpression('void', jsc.literal(0)));
```
rictic: We've been using `undefined` elsewhere, but we should be consistent either way.

I have a mild preference for `undefined` as it's clearer, and there's no way to redefine the toplevel undefined variable in an ES5 compliant browser. The only time `undefined` wouldn't be undefined is if you declared a scoped variable named `undefined`, there's no way for one file or function to change the meaning of `undefined` for another.

src/test/unit/project-converter_test.ts
```diff
@@ -975,6 +975,35 @@ export function arrowFn() {
...
+            window.NS = {
+              f: function() {
+                console.log('f', this);
+                this.g();
+              },
+              g: function() { console.log('g'); },
+            };
+          </script>`,
+      });
+      assertSources(await convert({namespaces: ['NS']}), {
```
rictic: The explicit namespaces argument shouldn't be necessary... I think.... since you've got the `@namespace` in the html.

e111077:
Also fixes #345 I was just dumb in describing the issue.

src/passes/rewrite-namespace-this-references.ts
```diff
@@ -25,7 +25,15 @@ function rewriteSingleScopeThisReferences(
     blockStatement: estree.BlockStatement, namespaceReference: string) {
   astTypes.visit(blockStatement, {
     visitThisExpression(path: NodePath<estree.ThisExpression>) {
-      path.replace(jsc.identifier(namespaceReference));
+      const parent = path.parent;
```
e111077: please change jsdoc for this function

src/passes/rewrite-namespace-this-references.ts
bicknellr: SGTM, updated.

src/test/unit/project-converter_test.ts
bicknellr: Yup, `namespaces` isn't needed here - removed. Although, removing the `@namespaces` annotation but leaving the `namespaces` argument doesn't work.

src/passes/rewrite-namespace-this-references.ts
bicknellr: Updated.

bicknellr:
@e111077, sorry, I didn't search for existing issues before filing that one.

bicknellr:
This case should really have a warning. If you're using this reference and running through modulizer, you're probably doing something wrong.

justinfagnani:
I'm not sure a warning is necessary... in the care we're targeting here it's fine.

src/passes/rewrite-namespace-this-references.ts
```diff
@@ -19,13 +19,23 @@ import * as jsc from 'jscodeshift';
...
+ * otherwise they are rewritten to `undefined`. Don't traverse deeper into new
+ * scopes.
  */
 function rewriteSingleScopeThisReferences(
     blockStatement: estree.BlockStatement, namespaceReference: string) {
   astTypes.visit(blockStatement, {
     visitThisExpression(path: NodePath<estree.ThisExpression>) {
-      path.replace(jsc.identifier(namespaceReference));
+      const parent = path.parent;
+      if (parent && parent.node.type !== 'MemberExpression') {
```
justinfagnani: Note, this won't work for `this['foo-bar']`, but that's probably _really_ rare.

bicknellr: I think it `MemberExpression` covers both `x.y` and `x['y']`: https://github.com/benjamn/ast-types/blob/2df3d43e3fdb736a74f4f2d40817f76beb4971f7/def/core.js#L306-L319

bicknellr: Oh, but following from that, what happens to `ns['x']` when `ns` is actually removed?

justinfagnani: Good to know. And... yeah, that won't work.

PR #418, opened 2018-05-05
Import path to meta

e111077:
Removed `--add-import-path` and changed it to `--remove-import-meta`. It might be useful to look at the individual commits to filter out all the tests

**before:**
The flag, `--add-import-meta`, that defaulted to `false`, would add
```js
class extends PolymerElement {
  static get importPath() { return import.meta.url; }
}

Polymer({
  importPath: import.meta.url
});
```
This was no longer supported in Polymer/polymer#5180

**after:**
The flag is now `--remove-import-meta` (`false` by default). So, by default we insert:
```js
class extends PolymerElement {
  static get importMeta() { return import.meta; }
}

Polymer({
  importMeta: import.meta
});
```
By adding the `--remove-import-meta` flag, the user can opt out of this insertion (which is now on by default).

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ Setting the import style allows you to set whether JavaScript imports are specif
 
 #### `--add-import-meta
 
-True by default, but if set to false, the static `importMeta` property will not be added to converted Polymer elements. See [the `importPath` documentation](https://www.polymer-project.org/2.0/docs/devguide/dom-template) for more information.
+True by default; the static `importMeta` property will be added to converted Polymer elements. See [the `importPath` documentation](https://www.polymer-project.org/2.0/docs/devguide/dom-template) for more information.
 
 
 ## Conversion Guidelines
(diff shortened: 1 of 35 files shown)


README.md
```diff
@@ -107,9 +107,9 @@ This will create a `modulizer_workspace` directory and checkout the repos and th
 
 Setting the import style allows you to set whether JavaScript imports are specified by npm package name, or relative file path. Importing specifiers that use package names are easier for third-party packages to work with, but unlike paths they currently can not run natively on the web. Defaults to "path".
 
-#### `--add-import-path`
+#### `--remove-import-meta
```
aomarks: I would lean towards keeping `--add-import-path` and just changing the default to `true`.

README.md
```diff
@@ -107,9 +107,9 @@ This will create a `modulizer_workspace` directory and checkout the repos and th
 
 Setting the import style allows you to set whether JavaScript imports are specified by npm package name, or relative file path. Importing specifiers that use package names are easier for third-party packages to work with, but unlike paths they currently can not run natively on the web. Defaults to "path".
 
-#### `--add-import-path`
+#### `--add-import-meta
 
-If included, the static `importPath` property will be added to converted Polymer elements. See [the `importPath` documentation](https://www.polymer-project.org/2.0/docs/devguide/dom-template) for more information.
+True by default, but if set to false, the static `importMeta` property will not be added to converted Polymer elements. See [the `importPath` documentation](https://www.polymer-project.org/2.0/docs/devguide/dom-template) for more information.
```
aomarks: I would just say "True by default. Add a static `importMeta` property to converted Polymer elements." or something (less negatives)

PR #426, opened 2018-05-11
Update yarn resolution for webcomponentsjs to 2.0.0

web-padawan:
This is still needed because of the [wct-browser-legacy](https://github.com/Polymer/tools/blob/master/packages/web-component-tester/wct-browser-legacy/package.json#L31) depending on the `^1.0.7` version. Releasing `wct-browser-legacy` 1.0.0 would be a perfect moment to bump that dependency. Still I think you have to cut a new release, and until that time we have to keep this resolution here.

/cc @usergenic

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-modulizer/426)

### DIFF
--- a/fixtures/packages/iron-icon/expected/package.json
+++ b/fixtures/packages/iron-icon/expected/package.json
@@ -31,7 +31,7 @@
     "samsam": "1.1.3",
     "supports-color": "3.1.2",
     "type-detect": "1.0.0",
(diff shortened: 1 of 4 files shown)


web-padawan:
Superseded by Polymer/tools#397

Will submit another PR to drop this stuff instead.

PR #430, opened 2018-05-11
Allow for running modulizer in a directory that is not git controlled

ergo:

### DIFF
--- a/src/cli/command-package.ts
+++ b/src/cli/command-package.ts
@@ -29,8 +29,14 @@ export default async function run(options: CliOptions) {
 
   // Ok, we're updating a package in a directory not under our control.
   // We need to be sure it's safe.
-  const {stdout, stderr} = await exec(inDir, 'git', ['status', '-s']);
-  if (!options.force && (stdout || stderr)) {
+  let stdout, stderr;
+  try {
+      ({stdout, stderr} = await exec(inDir, 'git', ['status', '-s']));
+  } catch (e) {
+      ({stdout, stderr} = e);
+  }
+  const isRepo = !((stderr || '').indexOf('Not a git repository') !== -1);
+  if (!options.force && isRepo && (stdout || stderr)) {
     console.error(
         `Git repo is dirty. Check all changes in to source control and ` +
         `then try again.`);


src/cli/command-package.ts
```diff
@@ -29,8 +29,18 @@ export default async function run(options: CliOptions) {
...
+      ({stdout, stderr} = await exec(inDir, 'git', ['status', '-s']));
+  } catch (e) {
+      ({stdout, stderr} = e);
+  }
+  const isRepo = !((stderr || '').indexOf('Not a git repository') !== -1);
+  if (!isRepo) {
+      console.warn(
+          `Not a git repo, proceeding.`);
+  }
+  if (!options.force && isRepo && (stdout || stderr)) {
```
stramel: I know we had talked about previously moving options.force up, so it won't even try to do a `git status`

src/cli/command-package.ts
```diff
@@ -29,8 +29,18 @@ export default async function run(options: CliOptions) {
 
   // Ok, we're updating a package in a directory not under our control.
   // We need to be sure it's safe.
-  const {stdout, stderr} = await exec(inDir, 'git', ['status', '-s']);
-  if (!options.force && (stdout || stderr)) {
+  let stdout, stderr;
+  try {
+      ({stdout, stderr} = await exec(inDir, 'git', ['status', '-s']));
+  } catch (e) {
+      ({stdout, stderr} = e);
```
justinfagnani: why this? What Errors have `stdout` and `stderr` on them?

src/cli/command-package.ts
justinfagnani: That seems good also, but you shouldn't need --force to run in a plain directory

src/cli/command-package.ts
ergo: https://github.com/Polymer/polymer-modulizer/blob/master/src/util.ts#L84
Normally when the Promise fails (when for example the dir is not controlled by git) it is unhandled and error object has those properties.

src/cli/command-package.ts
```diff
@@ -29,8 +29,18 @@ export default async function run(options: CliOptions) {
...
+  }
+  const isRepo = !((stderr || '').indexOf('Not a git repository') !== -1);
```
justinfagnani: Looks like double-negation here, can't this be:

```ts
const isRepo = (stderr || '').indexOf('Not a git repository') === -1;
```

In addition, we prefer explicit checks to boolean conversions, so:

```ts
const isRepo = stderr === undefined || stderr.indexOf('Not a git repository') === -1;
```

src/cli/command-package.ts
```diff
@@ -29,8 +29,19 @@ export default async function run(options: CliOptions) {
 
   // Ok, we're updating a package in a directory not under our control.
   // We need to be sure it's safe.
-  const {stdout, stderr} = await exec(inDir, 'git', ['status', '-s']);
-  if (!options.force && (stdout || stderr)) {
+  let stdout, stderr;
+  try {
+      ({stdout, stderr} = await exec(inDir, 'git', ['status', '-s']));
+  } catch (e) {
+    // Grab command execution results from exception info.
```
justinfagnani: Please format this file. `npm run format` will format the project, and then you can just commit this file.

src/cli/command-package.ts
```diff
@@ -29,8 +29,19 @@ export default async function run(options: CliOptions) {
...
+      ({stdout, stderr} = e);
+  }
+  const isRepo = !((stderr || '').indexOf('Not a git repository') !== -1);
+  if (!isRepo) {
```
justinfagnani: seems easier to reason about to move this code into the catch block above. Assume that if the command succeeded that `isRepo` is true.

PR #444, opened 2018-05-25
remove travis type and format checks

e111077:
this is so that we can include these in master but not in the modulzied files.


e111077:
wrong button!

PR #455, opened 2018-07-01
Add files property in package.json to reduce package size

web-padawan:
See for example the recently merged webcomponents/webcomponentsjs#929

@bicknellr please take a look if these files should be enough, or am I missing something?

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-modulizer/455)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,12 @@
   "bin": {
     "modulizer": "bin/modulizer.js"
   },
+  "files": [
+    "dependency-map.json",
+    "lib/",
+    "custom_typings/",
+    "bin/"
+  ],
   "scripts": {
     "build": "tsc",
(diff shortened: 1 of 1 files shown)


justinfagnani:
How much does this reduce the package size? The source shouldn't be that large.

web-padawan:
This is expected to ensure that no test fixturest etc get published occasionally.

bicknellr:
Should `src/` not also be in this list?

web-padawan:
I used the code from analyzer: https://github.com/Polymer/tools/blob/master/packages/analyzer/package.json

Feel free to modify this PR if you think `src` should be also published.

justinfagnani:
Opened https://github.com/Polymer/tools/pull/573 in the monorepo

PR #457, opened 2018-07-01
Do not set display: none for template elements

web-padawan:
Fixing the leftover from #411, really this should have been done there. These template elements are never added to the DOM, only their content is; so let's save a couple of bytes.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-modulizer/457)


web-padawan:
@justinfagnani @bicknellr PTAL

gentle ping to you guys, would be really nice to get my PRs merged and new modulizer released, so that we could use it to upgrade the versions of Vaadin components.

